PR #127, opened 2015-01-19
Update readme with skip link usage

gemmaleigh:
The template sets a skip link to `#content`. In order for this to work
`id=“content”` must be set on the main content area.

Adding @edds's reply to the Digital Service Designers mailing list to the readme.


edds:
Amazing. Thank you. This was on my todo list to do. Can you work through the rest of my todo list also ; )

PR #128, opened 2015-01-20
Remove inline JavaScript

alexmuller:
**Please don't merge this pull request in its current state. It's open for feedback only.**

We'd like to enable Content Security Policy on GOV.UK, which provides additional security against cross-site scripting and other vulnerabilities in the site.

As part of this, we need to set an HTTP header to instruct the browser to not execute any JavaScript that it encounters in `<script>` tags on the page.

This is to defend against cases where somebody exploits GOV.UK and we are not adequately protected against script tags injected in the page by other means eg by a malicious author in a content management system.

This commit moves inline JavaScript into asset files that can be loaded from a whitelisted domain.

This commit has issues and should **not** be merged to master. The issues I know of are:

- The ERB in `ie-8.js` is not processed
- This will cause an extra 3 HTTP requests, including a delay in adding the `js-enabled` class to the body, which will probably have negative side effects

I'd appreciate feedback on ways to fix those issues and any other thoughts from people.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,7 +10,7 @@ task :compile do
   Compiler::AssetCompiler.compile
 end
 
-desc "Build both gem and tar version"
+desc "Build all versions"
 task :build => ["build:gem", "build:tar", "build:play", "build:mustache", "build:liquid", "build:mustache_inheritance", "build:jinja", "build:ejs"]
 
 namespace :build do
--- a/lib/govuk_template/engine.rb
+++ b/lib/govuk_template/engine.rb
@@ -5,7 +5,10 @@ module GovukTemplate
         govuk-template*.css
         fonts*.css
         govuk-template.js
+        ie-8.js
+        ie-mobile-10.js
         ie.js
+        js-enabled.js
         vendor/goog/webfont-debug.js
       )
     end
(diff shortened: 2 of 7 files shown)


edds:
Is it possible to leave `js-enabled` and the IE 10 mobile be left inline. From [this spec](http://w3c.github.io/webappsec/specs/content-security-policy/) it says:

> Individual inline scripts and stylesheets may be whitelisted via nonces (as described in §4.2.4 Valid Nonces) and hashes (as described in §4.2.5 Valid Hashes).

Having two extra http requests for modern browsers will slow things down.

alexmuller:
I had a very long chat with @edds about this earlier (thanks @edds!) and here's a summary of where I think we got to:

- The inline JavaScript for IE8 doesn't need to be moved because IE8 doesn't support CSP
- The other two instances of inline JavaScript need to remain inline for performance reasons
- This repository should be responsible for creating a manifest of inline script hashes which can be used by the template consumer for setting a CSP header
- We'll then make changes in Slimmer to read in the manifest and set (or modify) a CSP header that includes the hashes from the manifest.

I'll repurpose this pull request to create a manifest of hashes.

PR #129, opened 2015-01-20
Use new grid layout placeholders for the cookie and header bar

gemmaleigh:
`inner-block` and `outer-block` have been deprecated, remove these
and replace with the new grid layout placeholders.

This affects the global header bar (the blue line underneath the
header) and the global cookie message (shown at first visit, or when js
is disabled).

Smaller screens:
govuk-template-mobile

Tablet:
govuk-template-tablet

Desktop:
govuk-template-desktop

### DIFF
--- a/source/assets/stylesheets/_header.scss
+++ b/source/assets/stylesheets/_header.scss
@@ -236,16 +236,11 @@
 /* Global header bar */
 
 #global-header-bar {
-  @include outer-block;
-  .inner-block {
-    @include inner-block;
-  }
-  .header-bar {
-    height: 10px;
-    background-color: $govuk-blue;
-    @include ie-lte(8) {
-      font-size: 0;
-    }
+  @extend %site-width-container;
+  height: 10px;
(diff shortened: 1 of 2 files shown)


edds:
This makes the code much nicer. Thanks for this @gemmaleigh!

dsingleton:
This is :+1:, long live the new grid.

Is there a risk of anyone hanging styles/behaviour off the removed selectors? Might be worth mentioning in release notes when we do.

gemmaleigh:
The only thing I can think of - `%site-width-container` styles are applied using the paragraph selector `#global-cookie-message p`  so if the cookie message was to have any other content, e.g. if `<%= yield :cookie_message %>`  was used and the content inside it wasn't a paragraph, that content wouldn't have the correct alignment.

I didn't want to start adding extra `inner` divs, but it might be to sensible to also wrap  `<%= yield :cookie_message %>`  in a `<p>` tag, but only if we expect this content to be a single paragraph.

PR #130, opened 2015-02-10
Add template location to "main" property in package.json

lennym:
It is useful when building modules which depend on the templates to be able to programmatically obtain the location of the template, especially so when installing as an npm package, as the install location may be dependent on other factors. Setting the "main" property of the package.json allows the full path of the tempalte to be ascertained using `require.resolve('govuk_template_*')` in a node environment.

See https://github.com/alphagov/govuk_template_mustache/issues/5


lennym:
Apologies if this breaks any tests, I tried looking through the tests and couldn't see anything it might obviously break, but I have almost zero ruby knowledge, so don't really know how to run said tests.

alexmuller:
If they're not running in Travis they don't exist. Looks good to me :+1:

PR #131, opened 2015-02-10
Update tests to reflect new package.json

alexmuller:
@alext pointed out that the tests for this run on our own internal CI, but only for branches that are from the alphagov repo.

Fix the tests for changes in #130.

We should consider whether we want to make builds run in Travis (as well?) in this repository, given that this repo is more likely to receive pull requests from external contributors.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -6,4 +6,9 @@ rm -f Gemfile.lock
 git clean -fdx
 bundle install --path "${HOME}/bundles/${JOB_NAME}"
 
-bundle exec rake spec build:and_release_if_updated
+bundle exec rake spec
+
+# If $PUBLISH_TEMPLATE is set, try and release the template
+if [[ -n "$PUBLISH_TEMPLATE" ]]; then
+  bundle exec rake build:and_release_if_updated
+fi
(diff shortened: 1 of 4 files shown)


lennym:
Thanks Alex!

tombye:
:+1:

alexmuller:
Tests are passing on Travis.

jenkins.sh
```diff
@@ -6,4 +6,9 @@ rm -f Gemfile.lock
 git clean -fdx
 bundle install --path "${HOME}/bundles/${JOB_NAME}"
 
-bundle exec rake spec build:and_release_if_updated
+bundle exec rake spec
+
+# If $TRAVIS is unset, try and release the template
+if [ -z "$TRAVIS" ]; then
+  bundle exec rake build:and_release_if_updated
+fi
```
alext: It would be better to make this only release if a variable is set - safer that way.  Then we can set the variable for the master Jenkins job.

This is how other gems that do branch builds work - eg: https://github.com/alphagov/gds-api-adapters/blob/master/jenkins.sh

alexmuller: Updated, thanks!

alexmuller: (and I've set the environment variable on CI)

PR #132, opened 2015-02-12
Update licence to correct format and date

tombye:
As per instructions from @annashipman, the copyright for this project should read "Crown Copyright", not "HM Government".

It is also 2015 now.

### DIFF
--- a/LICENCE.txt
+++ b/LICENCE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2013 GDS
+Copyright (c) 2015 Crown Copyright (Government Digital Service)
 
 MIT License
 


alext:
The copyright year should be the year the project started, not necessarily the current year.

tombye:
I was using this as my precedent: https://github.com/alphagov/govuk_frontend_toolkit/commit/429a85cf851f2788502b9227c1b3863cadb81fd4

@alext  Do you think it's now incorrect?

alext:
https://www.gov.uk/copyright would suggest that the year should be the of creation.  More information here: http://www.copyrightservice.co.uk/copyright/p03_copyright_notices

PR #134, opened 2015-02-16
Remove inner and outer block mixins

gemmaleigh:
- Replace `outer-block` with new grid layout placeholder
`%site-width-container`
- Remove `inner-block` as the left and right margins are now set by `%site-width-container`

When the template is compiled, the Sass warnings for the deprecated
grid mixins will no longer appear.

govuk_template_footer_tablet_nobg
govuk_template_footer_desktop
govuk_template_footer_mobile

### DIFF
--- a/source/assets/stylesheets/_footer.scss
+++ b/source/assets/stylesheets/_footer.scss
@@ -5,13 +5,13 @@
   border-top: 1px solid $footer-border-top;
 
   .footer-wrapper {
-    @include outer-block;
+    @extend %site-width-container;
+    background-color: $footer-background;
     padding-top: 20px;
+
     @include media(tablet) {
       padding-top: 60px;
     }
-    background-color: $footer-background;
-    margin: 0 auto;
   }
 
   a {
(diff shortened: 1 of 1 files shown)


edds:
This is a good change, and one that should happen. My only worry is that it changes the width of any content which is added to the template in `footer_top`. For example on GOV.UK we will need to adjust the width of our footer links block. Wondering how many other people use that space.

robinwhittleton:
`.footer-categories` on static needs to be updated to use grid columns before this can be merged (along with others?).

dsingleton:
I don't think `static` blocks this PR. Updating `.footer-categories` only needs to be done when the `govuk_template` version is bumped on static.

My only concern echo's Edds, that this may affect other consumers using the footer in a similar way to us, but I think this is fine if we bump to `0.18.0` and make the `CHANGELOG` clear (and cross link this discussion).

Any objections to me merging?

robinwhittleton:
None if you’re happy, go for it.

dsingleton:
:clap:

PR #135, opened 2015-02-24
Print visually hidden content

alextea:
This may be controversial, but content that is visually hidden on screen may be relevant for the printed version


bradleywright:
I really think this would be better applied on a case by case basis, rather than generally.

edds:
I agree with @bradleywright. I think it is to much of a may by relevant for us to have it set to print it always.

alextea:
OK, I'll close it :)

PR #138, opened 2015-03-02
Create NuGet package for ASP.NET MVC integration

daveraine:
Hi, I spoke with Tom Scott today about some work I've been doing to get the template and frontend toolkit integrated with ASP.NET MVC. He suggested I create a pull request so you can see what I have done.

I've created a new rake task, build:razor, which will create a nuget package (similar to a ruby gem) containing the files for easy install into an ASP.NET MVC website. I kept this separate in the rakelib folder, but this can easily be merged into your rakefile.

The razor_processor.rb is based off play_processor.rb. The template engine in Play is actually based of Razor which is used in ASP.NET MVC which made this an easy task. I inherited the razor_packager.rb from tar_packager.rb and added code to create a nuget package.

I've never written Ruby before so sorry if anything is a bit messy.

To create the nuget package nuget.exe is required, which I have placed in a "tools" folder. As this is a .NET exe (and the only way to create a nuget package), it requires mono to be installed on the machine. I have this working fine on an Ubuntu 14.10 machine. I wasn't sure on your exact development environment but assumed Linux, although this should work fine on OSX too.

One step I have not done is to publish this package to www.nuget.org. This is very similar to www.rubygems.org! If you are happy with what I have done then I will help with the publishing step. I didn't want to start publishing packages without approval.

I've modified the folder structure of what is inside the nuget package to better reflect common ASP.NET conventions. This puts all sass files in a "Content" folder, all javascript in a "Scripts" folder, and the main layout page in "Views/Shared". 

As I'm also interested in using the Frontend Toolkit, I've created a new repository to create a nuget package for that. It's based on how you are creating the gem and npm packages. Again, I've not got it to actually publish it yet, it only creates a local package, but this is an easy step to have once you have an account on www.nuget.org. You can see my new repository here https://github.com/daveraine/govuk_frontend_toolkit_nupkg. Again, I've modified the folder structure in the package for ASP.NET conventions.

Please let me know if you have any questions.

PR #139, opened 2015-03-10
ASP.NET MVC version of GOV.UK template

bradwright:
Replaces #138 

This is the scripts required to generate an ASP.NET MVC template package for the govuk_template. I've tidied up the Ruby and moved some things around.

Original PR contents:

> Hi, I spoke with Tom Scott today about some work I've been doing to get the template and frontend toolkit integrated with ASP.NET MVC. He suggested I create a pull request so you can see what I have done.

> I've created a new rake task, build:razor, which will create a nuget package (similar to a ruby gem) containing the files for easy install into an ASP.NET MVC website. I kept this separate in the rakelib folder, but this can easily be merged into your rakefile.

> The razor_processor.rb is based off play_processor.rb. The template engine in Play is actually based of Razor which is used in ASP.NET MVC which made this an easy task. I inherited the razor_packager.rb from tar_packager.rb and added code to create a nuget package.

> I've never written Ruby before so sorry if anything is a bit messy.

> To create the nuget package nuget.exe is required, which I have placed in a "tools" folder. As this is a .NET exe (and the only way to create a nuget package), it requires mono to be installed on the machine. I have this working fine on an Ubuntu 14.10 machine. I wasn't sure on your exact development environment but assumed Linux, although this should work fine on OSX too.

> One step I have not done is to publish this package to www.nuget.org. This is very similar to www.rubygems.org! If you are happy with what I have done then I will help with the publishing step. I didn't want to start publishing packages without approval.

> I've modified the folder structure of what is inside the nuget package to better reflect common ASP.NET conventions. This puts all sass files in a "Content" folder, all javascript in a "Scripts" folder, and the main layout page in "Views/Shared".

> As I'm also interested in using the Frontend Toolkit, I've created a new repository to create a nuget package for that. It's based on how you are creating the gem and npm packages. Again, I've not got it to actually publish it yet, it only creates a local package, but this is an easy step to have once you have an account on www.nuget.org. You can see my new repository here https://github.com/daveraine/govuk_frontend_toolkit_nupkg. Again, I've modified the folder structure in the package for ASP.NET conventions.

> Please let me know if you have any questions.

Still todo:

- [x] Verify this approach generates the correct package
- [ ] Add tests
- [ ] Install Mono on CI tools
- [ ] Get this to upload the package to Nuget automatically


bradleywright:
@daveraine can you verify that this still does what you expect?

build_tools/packager/razor_packager.rb
```diff
@@ -0,0 +1,53 @@
...
+      target_dir.mkpath
+      target_file = File.basename(file, File.extname(file)).sub(/\.html\z/, '.cshtml') # /path/to/foo.html.erb -> foo.cshtml
+      File.open(target_dir.join(target_file), 'wb') do |f|
+        f.write Compiler::RazorProcessor.new(file).process
+      end
+    end
+
+    def prepare_contents
+      super
+      run "mono #{@repo_root}/tools/nuget.exe pack GovUK.Template.Razor.nuspec -Version #{GovukTemplate::VERSION}"
```
daveraine: The version is being set by the -Version command line parameter here

source/GovUK.Template.Razor.nuspec.erb
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+    <metadata>
+        <id>GovUK.Template.Razor</id>
+        <version><%= GovukTemplate::VERSION %></version>
```
daveraine: Setting the version from the command line in razor_packager.rb will override whatever is in here (even though they are getting set to the same!)

bradleywright: I think it's more explicit to set it in the package definition, so I'll remove the command line flag.

daveraine:
Hi, looking good and still working fine. Just a comment on setting the version, there isn't a need to do it in both the .nuspec file and from the command line call. I'd done it from the command line as it was easier than feeding the version into the .nuspec file.

Also, regarding the nuget.exe download, the latest version can always be downloaded from here: http://nuget.org/nuget.exe. Also calling

````
nuget.exe update -self
````

will get the exe to update itself.

bradleywright:
@daveraine okay I've updated this again. Could you take another look? (Sorry about the ask - no-one here knows much about ASP.NET so we need help!)

daveraine:
No problem, happy to help.

All working correctly.

jamiecobbett:
I've rebased onto master made some small changes to make this mergeable from a git point of view.

I think the `.travis.yml` should have the mono (and nuget?) dependencies added (some information here: http://docs.travis-ci.com/user/languages/csharp/), as will our internal CI puppet.

daveraine:
Only the Mono runtime is required, nothing is getting built in C#, so I'm not sure if you will need to add Mono to your travis configuration as I'd imagine that would be the Mono SDK. However, I'm unfamiliar with travis, and that might be the only way to get the Mono runtime as a dependency.

jamiecobbett:
@daveraine thanks, I was mistaken about Travis - it doesn't try to test-build the packages (though I think it should).

bradleywright:
I think `apt-get install mono-devel` should give you the CLI `mono` command.

daveraine:
That will give the full SDK, I think `apt-get install mono-runtime` will be enough.

daveraine:
I've found an issue with the template generation. I've rebased my own copy of this branch from master and made the update in razor_processor.rb. You can see the updates here:

https://github.com/daveraine/govuk_template/blob/update-dotnet-mvc-nuget-package/build_tools/compiler/razor_processor.rb

The update is to use `@Html.Raw()` to output the strings, as if they contain html they are not output correctly.

Not sure the best way about getting this update to you, I'm not sure if I can do  pull request to a pull request if it has been rebased?

Davidslv:
This PR has been here for almost a year, is it still relevant? Would someone care to check this please?

dsingleton:
I'm going to close this for now - if someone needs this version of the template, please let us know and we'll pick this back up.

In the mean time i've rebased the branch, fixed conflicts and added @daveraine's [commit](https://github.com/alphagov/govuk_template/pull/139/commits/c44184114b1d8c36dba3bd638589d7ae95d2f8fc) - so the unit tests pass and the branch is in a state someone can pickup more easily.

I also had a [very quick go](https://github.com/alphagov/govuk_template/pull/139/commits/99dedf5c18a038dc5123aedac4bdec496c915060) at getting `mono` working in Travis, which didn't work.



As

daveraine:
Thanks @dsingleton for moving my update over. As you say, this branch is in a state that someone can pick it up and carry on with it if needed.

PR #140, opened 2015-03-19
Add webjar packager, spec and update README.md

ijmad:
Hello, this is some code to build the assets in to a WebJar. I'd like this for my project. Let me know if my Ruby is OK because I write it poorly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The Ruby language (1.9.3+), the build tool [Rake](http://rake.rubyforge.org/) & 
 
 ## Packaging
 
-At present this generates 8 output formats:
+At present this generates 9 output formats:
 
 1. a gem containing a Rails engine
 2. a tarball containing Play Framework templates
@@ -24,6 +24,7 @@ At present this generates 8 output formats:
 6. a tarball containing Jinja templates
 7. a tarball
 8. a tarball containing EJS (Embedded JavaScript) templates
+9. a JAR file containing assets structured as per [WebJars](http://www.webjars.org/)
 
 ### Gem version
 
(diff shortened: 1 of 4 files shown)


bradleywright:
Awesome, thanks @ijmad.

Do we need to push the WebJar anywhere, or are you intending to use this library directly in your build process?

ijmad:
We do need to host it somewhere, but it needs to go somewhere that looks like a Maven/Ivy repository, and we currently don't have a Nexus or similar. The structure of these repos isn't a great match for github's release manager.

It is possible though, via some mild hacking, to do this with a Maven plugin that GitHub provides, which I've been doing over here - https://github.com/alphagov/govuk_template_jar/ (see https://github.com/alphagov/govuk_template_jar/tree/mvn-repo for the fake repo).

I could fold this in to govuk_template as well, but it is a bit messy.
General thoughts welcome, and/or a better option if you can think of one.

bradleywright:
I just wanted to check if you had any thoughts about hosting before I merged this - I've tested it on a GOV.UK Vagrant machine and it builds correctly, so I think it's good to merge.

We can revisit hosting of the JAR later.

bradleywright:
:+1:

ijmad:
No worries - we can have a chat about that at some point!

jamiecobbett:
When I build the webjar, it doesn't seem to include the view (`govuk_template.html`). Is this deliberate?

jamiecobbett:
Spoke with @ijmad in person, and yes it is deliberate because webjars aren't for templates/html. I've tweaked the README to make this clearer over in https://github.com/alphagov/govuk_template/pull/146

PR #141, opened 2015-03-26
Abstract all strings

froddd:
There were still a few strings/messages left in the main template that were hard-coded: these were posing a problem for creating translated (Welsh) versions of services, which we've just started doing for our service (Make a Plea).

I've abstracted these remaining strings into several new placeholders, and updated the compilers to suit. I've made sure these placeholders are optional and all default to the original string, to maximise compatibility.

The Play compiler may need looking at, as I have the feeling some extra references to the new blocks need to be added here and there. 

Will these additions require new tests? Will the publishers and packagers need to be updated?

Still to do:
- [ ] Check Play compiler
- [ ] Update/add tests if necessary?
- [ ] Check packagers/publishers


tombye:
Nice work @froddd. @michaeldfallen would you be able to check the Play compiler changes?

The packages build ok so the publishers should work. @edds @dsingleton Thoughts?

froddd:
Hi guys, is there any progress on checking this? This is quite important for us to move forward -- we're currently having to use a personal forked repo for some of our requirements until this is merged; if we could get this into master that would help remove such dependencies. Cheers!

bradleywright:
Assuming that the Play compiler still works I think we should merge this.

Also, looking at the diff (where missing strings were added to a bunch of compilers) we should probably add a test that the compilers have parity with their string coverage.

jamiecobbett:
> Update/add tests if necessary?

> Also, looking at the diff (where missing strings were added to a bunch of compilers) we should probably add a test that the compilers have parity with their string coverage.

I have made an attempt at a test which checks this: https://github.com/alphagov/govuk_template/pull/146

bradleywright:
Hi @froddd - we've just merged #146, which introduces template assertions. Could you please rebase against master and check the tests pass?

froddd:
Hey @bradleywright, I've now merged master back into this. I've had to update a couple of things after merging, but tests are passing so all should be good!

I've also noticed the defaults (if any) are not that consistent for each block/tag between the different compilers, but this should probably be left for another PR altogether.

build_tools/compiler/play_processor.rb
```diff
@@ -12,7 +12,7 @@ class PlayProcessor < TemplateProcessor
       # top_of_page has a special purpose: it is required by Play to define the
       # parameters to pass when rendering
       # https://www.playframework.com/documentation/2.2.x/ScalaTemplates#Template-parameters
-      top_of_page: '@(title: Option[String], bodyClasses: Option[String], htmlLang: Option[String] = None)(head:Html, bodyStart:Html, bodyEnd:Html, insideHeader:Html, afterHeader:Html, footerTop:Html, footerLinks:Html, headerClass:Html = Html.empty, propositionHeader:Html = Html.empty, homepageUrl:Html = Html.empty, globalHeaderText:Html = Html.empty, cookieMessage:Html = Html.empty)(content:Html)',
+      top_of_page: '@(title: Option[String], bodyClasses: Option[String], htmlLang: Option[String] = None)(head:Html, bodyStart:Html, bodyEnd:Html, insideHeader:Html, afterHeader:Html, footerTop:Html, footerLinks:Html, headerClass:Html = Html.empty, propositionHeader:Html = Html.empty, homepageUrl:Html = Html.empty, globalHeaderText:Html = Html.empty, cookieMessage:Html = Html.empty, skipLinkMessage:String, logoLinkTitle:String, licenceMessage:Html, crownCopyrightMessage:String)(content:Html)',
```
jamiecobbett: I think `crownCopyrightMessage:String` should be `crownCopyrightMessage:Html`. I'm guessing the difference in type is used to decide whether to escape any HTML or not, so as it stands it would lead to (the default value) being escaped.

jamiecobbett:
I think these new parameters should be added to the integration tests. I've made an attempt here: https://github.com/alphagov/govuk_template/commit/10544133b75f1bcd01bb109aac86e238c8bb9db3

I will add something to the README that when adding fields, they need to be added to the integration tests, or make the integration tests fail if they aren't added (trickier).

build_tools/compiler/play_processor.rb
froddd: Is this in respects to the HTML entity in the default string? Will this affect any value passed as the variable, and if so, should we extend the reasoning to the other strings, to allow for special characters?

jamiecobbett: @froddd 

> Is this in respects to the HTML entity in the default string? 

Yes.

> Will this affect any value passed as the variable, and if so, should we extend the reasoning to the other strings, to allow for special characters?

I based my comment on the other parameters. I'm afraid I don't know enough about Play templates to say with any confidence - maybe @HolyHaddock could explain?

HolyHaddock: As @jamiecobbett  suggests, values with a String type will be escaped before being rendered, Html won't.
If the variable might contain HTML, and definitely won't include unsanitised external values, Html might be more appropriate for the other values too.

froddd:
Thanks @jamiecobbett -- I've merged these.

Also changed type to Html for all new strings: only one of the defaults contains a special character but I guess we probably want to be on the safe side!

PR #144, opened 2015-05-05
added pip version to readme

Phemus:
pip version has been created on pypi added documentation for it to readme


jamiecobbett:
Hi @Phemus, thanks for suggesting this. Unfortunately the version [published on pypi](https://pypi.python.org/pypi/govuk-template/0.2.41) is not built from this library. It is instead built from the [LandRegistry library](https://github.com/LandRegistry/python-govuk-template), which has a point-in-time copy of our Mustache template and is versioned separately. Therefore, I think you should this documentation to that repo, not this one.

PR #145, opened 2015-05-13
Support custom homepage URLs for header link

jamiecobbett:
In every environment, this link takes you to www.gov.uk. This seems unhelpful
when you're in an environment that isn't production as it takes you to a
different environment which can be hard to notice.

This already happens in preview, and now that we are adding GOV.UK Draft (which
is accessed on a different domain), there will be more users looking at the
frontend of GOV.UK in an environment that isn't on the www.gov.uk domain.

This will be paired with a change to static which will use Plek to get the
current website root and pass it in.

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -81,7 +81,7 @@
       <div class="header-wrapper">
         <div class="header-global">
           <div class="header-logo">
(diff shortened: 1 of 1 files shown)


lennym:
:+1:

jamiecobbett:
I think I need to add lines to each of the compilers in `@@yield_hash`?

jamiecobbett:
Closing until https://github.com/alphagov/govuk_template/pull/146 has been merged/rejected.

PR #146, opened 2015-05-13
Make compilers more robust

jamiecobbett:
Some refactoring, but most significantly, I've added a test to make it harder to add calls to `yield` to the ERB without them being supported in the other languages and fixed some bugs around that.

### DIFF
--- a/build_tools/compiler/play_processor.rb
+++ b/build_tools/compiler/play_processor.rb
@@ -4,6 +4,24 @@ require 'compiler/template_processor'
 module Compiler
   class PlayProcessor < TemplateProcessor
 
+    @@yield_hash = {
+      layout: "<!-- Page content goes here -->",
+      html_lang: "@htmlLang.getOrElse(\"en\")",
+      page_title: "@title.getOrElse(\"GOV.UK - The best place to find government services and information\")",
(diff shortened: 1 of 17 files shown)


build_tools/compiler/play_processor.rb
```diff
@@ -1,9 +1,28 @@
...
+      header_class: '@headerClass',
+      proposition_header: '@propositionHeader',
+      content: '@content',
+      body_start: '@bodyStart',
+      body_end: '@bodyEnd',
+      inside_header: '@insideHeader',
+      after_header: '@afterHeader',
+      footer_top: '@footerTop',
+      footer_support_links: '@footerLinks',
+      cookie_message: '@cookieMessage.getOrElse("")',
```
jamiecobbett: This needs to have the HTML fragment that is used as the runtime default in the ERB template, otherwise all Play clients have to supply their own cookie warning.

jamiecobbett: Fixed and rebased.

bradleywright:
Looks good apart from https://travis-ci.org/alphagov/govuk_template/builds/62537279 failures.

jamiecobbett:
I think there's one more thing to solve - checking that the PlayProcessor `top_of_page` key is (vaguely) correct - I think the best way is to test that every other key appears at least once in it.

jamiecobbett:
@bradleywright I think this is now good to merge.

bradleywright:
Code looks good - as discussed IRL it'd be nice to give the compiled templates a once over vs master to see if they differ much (mostly for Scala).

jamiecobbett:
Checked the diff for the Scala/Play template when built on master versus on this branch and it looks valid in my Scala syntax highlighter.

PR #148, opened 2015-05-14
Support custom global header text

jamiecobbett:
We want to be able to override the "GOV.UK" in the top left of the header with
"GOV.UK (Draft)" or similar to clearly signal when a user (eg a content editor)
is looking at Draft GOV.UK.


jamiecobbett:
Rebased onto master. The Travis build can take quite a long time before starting. /cc @dsingleton @edds

jamiecobbett:
:fireworks:

PR #149, opened 2015-05-15
Integration tests

jamiecobbett:
We want to be confident that the changes we make to govuk_template don't
break the integrations with the various languages/frameworks. The only way we
can really do this automatically is to have dummy apps in each of the languages. If we have an automated test suite, we can confidently merge pull requests.


bradleywright:
Random thought: this seems like an ideal use case for Docker.

jamiecobbett:
Added a Mustache test app.

jamiecobbett:
Added a Node.js Mustache inheritance test app.

dsingleton:
I think integration tests are a ver good idea :clap:

However, I think including those tests within the main repo is going to get unwieldily and could cause confusion, especially if examples are added for all supported templating formats and languages. One option could be to split the integration tests out into a new repo pulled in (ala content-schemas), or a series of dependant projects (ala content-schemas in a different way)?

I'm also not sure we'd need full app-level integration, it's bulkier and conflates template language vs app language. Eg, Mustache templates can be used in a number of different apps/languages, and I'm not sure we'd want to tests distinct ruby, python, JS (web), JS (server) apps in those situations.

Semantically we want be testing the template against a trustworthy implementation of that templating format, which should give us a huge increase in confidence when making changes. There may be some cases where cross-language/in-app testing of formats is useful, but it feels like a problem to deal with later on.

So i'm well in favour of the idea here, but I think the approach needs a bit more discussion

jamiecobbett:
@dsingleton thank you for the feedback.

> However, I think including those tests within the main repo is going to get unwieldily and could cause confusion, especially if examples are added for all supported templating formats and languages. One option could be to split the integration tests out into a new repo pulled in (ala content-schemas), or a series of dependant projects (ala content-schemas in a different way)?

If we did split it into a separate repo, I think we'd have problems with workflow - you would need to merge the branch of the tests repo to have the branch build of `govuk_template` be able to pass, at which point the master build of `govuk_template` will start failing and you're not even sure the branch of `govuk_template` will get merged.

> I'm also not sure we'd need full app-level integration, it's bulkier and conflates template language vs app language. 

I agree full app-level integration is unnecessary. The only place where I've done this is for Rails, because the `.erb` template we distribute is really-Rails specific - it uses `content_for?()` and `asset_path()` which are part of Rails, not ERB. There may be a way of using ActionView directly to render the template.

> Eg, Mustache templates can be used in a number of different apps/languages, and I'm not sure we'd want to tests distinct ruby, python, JS (web), JS (server) apps in those situations. Semantically we want be testing the template against a trustworthy implementation of that templating format, which should give us a huge increase in confidence when making changes. There may be some cases where cross-language/in-app testing of formats is useful, but it feels like a problem to deal with later on.

Agreed - I don't think I've done anything contrary to that, but perhaps it is worth documenting.

dsingleton:
> Agreed - I don't think I've done anything contrary to that, but perhaps it is worth documenting.

Sorry, I looked at the Rails example and assumed the others were full apps. Agreed that if we could pull in rails as a dependency and just use ActionView, rather than having a whole app, that'd be a lot better.

integration_tests/README.md
```diff
@@ -0,0 +1,19 @@
+# Integration Tests
```
dsingleton: I think we could avoid the term "app" and just use integration/implementation test. Also worth linking from the main `README`, as you mentioned.

Otherwise, this is good documentation and gets across the intent of the tests really well.

dsingleton:
Had a quick chat in person. @jamiecobbett agreed to investigate a more minimal implementation of the rails integration test, to avoid setting precedent of full app implementation.

That said, if thats isn't straight forward, i wouldn't want to block this PR for that, though i'd want to revisit it later :)

jamiecobbett:
I think this is now ready to merge - @edds would you mind giving it a once over?

bradleywright:
This has changes that will conflict on master (although it appears to be able to merge cleanly?).

Can you confirm @jamiecobbett ?

jamiecobbett:
@bradleywright I've rebased onto master to make sure - looks good to me.

PR #150, opened 2015-05-21
Add missing closing tag

colmjude:

PR #152, opened 2015-05-22
Use correct Jinja tag for homepage url

jamiecobbett:
Following on from: https://github.com/alphagov/govuk_template/pull/150#issuecomment-104367421


dsingleton:
https://github.com/alphagov/govuk_template/pull/153 for the above comment

PR #161, opened 2015-07-06
Fix Jenkins CI build

jamiecobbett:
The two integration apps were running bundler without specifying a path to
install to, which meant that bundler was installing gems system-wide. On our
Jenkins CI, the user doesn't have privileges to install system-wide, so the
build was failing.

Thanks to @alext for doing the detective work.

### DIFF
--- a/integration_tests/integrations/rails/.gitignore
+++ b/integration_tests/integrations/rails/.gitignore
@@ -1,0 +1,1 @@
+vendor/bundle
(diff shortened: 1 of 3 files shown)


integration_tests/integrations/rails/.gitignore
```diff
@@ -0,0 +1 @@
+vendor/bundle
```
alext: Do you need a similar `.gitignore` for the mustache one as well?

jamiecobbett: No... and I don't understand why not.

alext: Ah, it already has a vendor directory that's ignored - https://github.com/alphagov/govuk_template/blob/master/integration_tests/integrations/mustache/vendor/.gitignore

PR #162, opened 2015-07-10
Django processor

froddd:
Django compiler and packager (including tests).

The Django compiler is quite similar to the Jinja one, with the difference being mainly the addition of the `staticfiles` wrapper: this means loading the `staticfiles` dependency in `top_of_page`; it also means the assets don't need the version appended, as this is handled by the wrapper at render time if necessary (we use hash-based cache-busting on our service).

Done:
- [x] Create compiler
- [x] Create packager
- [x] Write tests

To do:
- [ ] Create publisher (separate git repo or pypi.python.org package?)


spec/build_tools/packager/django_packager_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+        generated_template = File.read(generated_template_path)
+        expect(generated_template).to match(%r[\A{% load staticfiles %}{% block top_of_page %}{% endblock %}])
+        expect(generated_template).to match(%r[href="{% static 'stylesheets/govuk-template\.css' %}"])
+      end
+
+    end
+  end
+
+end
```
jamiecobbett: Nitpick: valid Unix files should have a trailing newline.

froddd: Agreed, sorted ;)

jamiecobbett:
One thing we'd really like to see for each new template that we provide is to have an integration test. There's a README which (hopefully) explains this: https://github.com/alphagov/govuk_template/blob/master/integration_tests/README.md

Is this something that you'd be prepared to do?

froddd:
Yeah, I'm happy to have a go at that.

froddd:
Here's the integration test, which has already highlighted one issue (now fixed!) with the template tag default filter syntax :)

I've had to use `sudo` in the `build.sh` script to install Django -- is that ok?

bradleywright:
Is there any way you could do this in a `virtualenv` instead? If Travis does it automatically as you say in 380a121, it should work fine.

froddd:
Looks like using `sudo` might be the only option here. If Travis was set up for Python we might have access to `virtualenv`, but in this case we don't automatically -- not really an option anyway, seeing as the core code is Ruby!

The only way we could force Django to be installed in a virtualenv seems to require sudo-ing the install of `virtualenv` itself, then instantiating the environment so we can run `pip install Django` without sudo -- this would just add more complexity IMO.

Travis seems to run completely isolated build machines anyway (http://docs.travis-ci.com/user/ci-environment/#Overview), so is this a huge problem? If no other suggestion I'll revert to using `sudo pip install Django`.

Any further input appreciated -- these things are not really what I'm best at, I feel like I'm stabbing in the dark a little here! :-/

integration_tests/integrations/django/build.py
```diff
@@ -0,0 +1,26 @@
...
+    "html_lang": "rb",
+    "skip_link_message": "Custom skip link text",
+    "logo_link_title": "Custom logo link title text",
+    "crown_copyright_message": "Custom crown copyright message text",
+}
+
+content = render_to_string("test_template.html", context)
+
+with open("../../html_for_testing/django_integration_test_app.html", "w") as static_file:
+    static_file.write(content)
```
jamiecobbett: Nitpick: needs a trailing newline.

integration_tests/integrations/django/.gitignore
```diff
@@ -0,0 +1,2 @@
+.env
+vendor/*
```
jamiecobbett: Nitpick: needs a trailing newline.

integration_tests/integrations/django/templates/test_template.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% block content %}The page content{% endblock content %}
+
+{% block footer_top %}<footer-top></footer-top>{% endblock footer_top %}
+
+{% block footer_support_links %}<footer-support-links></footer-support-links>{% endblock footer_support_links %}
+
+{% block licence_message %}Custom license message text{% endblock licence_message %}
+
+{% block body_end %}<body-end></body-end>{% endblock body_end %}
```
jamiecobbett: Nitpick: needs a trailing newline.

froddd:
Is this good to go, or do we require a publisher be put in place before this can be merged?

If so, as per question in original PR: should this publish to git repo (which needs creating -- I don't have write access on alphagov), or publish as Pypi package?

jamiecobbett:
@froddd thanks for doing the work on this - apologies for the delay in getting it merged.

With regard to adding a publisher: when the version is next bumped, a `tgz` will be uploaded to https://github.com/alphagov/govuk_template/releases. If you need it to be published elsewhere then, someone would need to add a publisher class to do that.

jamiecobbett:
@froddd this will be released as v0.16.0 once this is merged: https://github.com/alphagov/govuk_template/pull/178

PR #163, opened 2015-07-10
Revert logo text override.

elliotcm:
This was added to distinguish draft frontends from
live, but we're adding a large "DRAFT" watermark
which will do the job instead.

https://trello.com/c/GCp2o5SA/271-add-draft-watermarking-to-content-preview-items


jamiecobbett:
The pain I went through :crying_cat_face:

elliotcm:
@jamiecobbett I know :(

quis:
This would still have been a nice feature for stuff like prototypes and styleguides which don't live on GOV.UK domains. Would anyone object to me reverting the revert?

PR #164, opened 2015-07-13
Don't escape output of statement_tag_for (Jinja)

quis:
The Jinja template tags generated by this compiler escape any variables passed to them.

So when you pass a string like `&copy; Crown Copyright` what gets spat out into the source of the page is `&amp;copy; Crown Copyright`, which in a browser looks like:

image

This commit adds a [Jinja filter called 'safe'](http://jinja.pocoo.org/docs/dev/templates/#working-with-automatic-escaping) which prevents escaping of special characters.

### DIFF
--- a/build_tools/compiler/jinja_processor.rb
+++ b/build_tools/compiler/jinja_processor.rb
@@ -9,7 +9,7 @@ module Compiler
     end
 
     def self.statement_tag_for(key, default_value)
-      "{{ #{key}|default('#{default_value}') }}"
+      "{{ #{key}|default('#{default_value}')|safe }}"
     end
 
     @@yield_hash = {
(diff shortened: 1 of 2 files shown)


edds:
LGTM but not sure of the implications of adding a `safe` filter to all statement tags.

quis:
@edds I'm not entirely sure of the implications either. d4c0925 is probably a better approach.

bradleywright:
Would this be better solved with a literal copyright character? eg `©`

quis:
@bradleywright 
1. That would require anyone passing a custom (eg localised) copyright message to know that it only accepts literal characters 
2. The default value in the other template compilers is `&copy;`; is there a reason the Jinja one should be different?

bradleywright:
Fair enough. Reason I ask is that we're adding some special cases just for one particular entry, so it seems like Unicode might be a quick answer without risking any leakage of unescaped content.

tombye:
Based on the comments, which seem to agree this can go in, I think this is ready for merge.

PR #165, opened 2015-07-20
Release 0.14.1

tombye:
This brings in a set of fixes for various issues hence the patch-level increment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # 0.14.1
 
 - Stop copyright entity being escaped in Jinja template
-- Update Crown Copyright link to new URL
 
 # 0.14.0
 
+- Update Crown Copyright link to new URL
 - Update the font-files and move woffs into CSS
 - All user-visible text in the template can now be overridden. Some template languages which don't support default values (such as plain mustache) will therefore lose that text and require you to handle supplying the default.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.14.1
+
+- Stop copyright entity being escaped in Jinja template
+- Update Crown Copyright link to new URL
+
 # 0.14.0
 
 - Update the font-files and move woffs into CSS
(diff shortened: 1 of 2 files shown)


quis:
https://github.com/alphagov/govuk_template/compare/v0.14.0%E2%80%A6b05e385

dsingleton:
It's a bit late now, but +1 to reverting https://github.com/alphagov/govuk_template/pull/163.

If we do remove it then the changelog should be a bit more explicit, so anyone using it knows they should remove the override value they're setting. (i think it's unlikely anyone is using it, but we should act in good faith here).

quis:
@dsingleton It's being used here, at least as an example: https://github.com/alphagov/govuk_prototype_kit/pull/15

dsingleton:
> @dsingleton It's being used here, at least as an example: alphagov/govuk_prototype_kit#15

Then +2 to keeping it :) It'd be useful for things like the component-guide and govspeak-preview too.

quis:
@dsingleton https://github.com/alphagov/govuk_template/pull/166

dsingleton:
Merged #166, so the diff against master now looks like: https://github.com/alphagov/govuk_template/compare/v0.14.0%E2%80%A6master

quis:
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.14.1
+
+- update Crown Copyright link to new URL
+- fix Jenkins CI build by adding a path to install gems to
```
dsingleton: I think we can drop this, it's not helpful to users of the gem

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+- revert override for GOV.UK text in logo
```
dsingleton: We can drop this now the revert has been reverted

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.14.1
+
+- update Crown Copyright link to new URL
```
dsingleton: Looking at the diff between the last version and master (https://github.com/alphagov/govuk_template/compare/v0.14.0%E2%80%A6master) i don't see the link change?

dsingleton:
:+1: thank you for the persistence!

tombye:
Thanks for the help :)

quis:
:fireworks:

PR #167, opened 2015-08-10
Show a black-on-transparent logo when printing

robinwhittleton:
Currently when printing a document that uses the govuk_template the logo appears as white-on-black next to black-on-transparent text. This PR fixes it so that the black-on-transparent logo is used for printing instead.

The basic approach:
1. Use the print black-on-transparent logo in the HTML
2. Hide it with the screen CSS and display the white-on-black logo as a CSS background image

This PR also tidies a couple of nits and shrinks the crown logos a (very) little.


russellthorn:
:+1: Looks good to me. Although I've not cloned this and viewed it locally.

edds:
Do we need this file? `source/assets/images/gov.uk_logotype_crown-1x.png` As far as I can tell the stylesheet uses the other copy and this would never be called.

edds:
Also is `source/assets/images/gov.uk_logotype_crown.png` used anymore now you no longer reference it from `source/views/layouts/govuk_template.html.erb` and can it be deleted?

robinwhittleton:
Yes, I guess those can both be removed.

PR #169, opened 2015-08-17
👻 Move inside_header to stop being invisible 🙈

roc:
- When adding e.g. 'GOV.UK' as inner header content we were getting a hidden piece of content and no link
- This should put that text into the anchor link and make it visible again

## Before
screen shot 2015-08-17 at 16 04 05

## After
screen shot 2015-08-17 at 16 03 47


edds:
You are supposed to use 'global_header_text` for the GOV.UK text. `inside_header` can be used for things like the search box.

PR #174, opened 2015-08-26
The copyright logo is safe

edds:
Otherwise the `&` on the `&copy;` gets turned into `&amp;` which means
the footer reads as "&copy; Crown copyright"


robinwhittleton:
Is there a reason why we’re using &amp; and not just © in the source?

PR #177, opened 2015-09-16
Fix extended footer on certain pages

robinwhittleton:
Some pages (e.g. https://www.gov.uk/government/publications/armed-forces-corporate-covenant-signed-pledges) have a long revision history. This, by default, has a class of `visuallyhidden` that hides the extended content off screen using `position:absolute` and `left:-9999em`. While this is hidden, the viewport will extend itself to include absolutely positioned elements, even if they’re positioned horizontally off screen.

This patch clips off screen elements so that they don’t take up any vertical space. It doesn’t affect VoiceOver’s ability to read the extended content.

### DIFF
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -25,7 +25,9 @@
 .visuallyhidden {
   position: absolute;
   left: -9999em;
-  height: 0;
+  top: auto;
+  width: 1px;
+  height: 1px;
   overflow: hidden;
 
   /*
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -25,6 +25,8 @@
 .visuallyhidden {
   position: absolute;
   left: -9999em;
+  height: 0;
+  overflow: hidden;
 
   /*
    * Extends the .visuallyhidden class to allow the element to be


tombye:
Looks correct to me. If we want non-visual content to occupy vertical space we have [a class for that](https://github.com/alphagov/govuk_template/blob/large_footer_fix/source/assets/stylesheets/_accessibility.scss#L44).

:+1:

edds:
I seem to have a hazy memory that adding a high of zero and overflow of hidden stops some content from being announced in some screen readers as the content is deemed to not be on the page. However I can't remember where I have that memory from, @alicebartlett can you remember?

edds:
Though I remembered it being a thing, here it is: https://github.com/alphagov/finder-frontend/commit/a16becb

robinwhittleton:
Interesting that @alicebartlett ’s commit message mentions Voiceover, as that’s what I tested in. Newer version fixes this maybe? Either way, setting a small height instead of a zero height would accomplish the same thing and might fix screen reader issues. I’ll amend the patch and test.

robinwhittleton:
Or maybe @aduggin knows off the top of his head? :)

fofr:
Has there been any movement on this since September?

robinwhittleton:
Nope. I really ought to find time to test it in Voiceover and JAWS.

jackscotti:
Any update on this?

tvararu:
For anybody testing this, you can use this document as a test case:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Test</title>
    <style>
      .hidden {
        height: 0;
        overflow: hidden;
      }
    </style>
  </head>
  <body>
    <div>Content</div>
    <div class="hidden">Hidden content</div>
  </body>
</html>
```

It seems to work correctly on OSX in voiceover in Chrome and Safari.

accessiblewebuk:
The test case by @tvararu works fine for me using the following combinations:
Windows 10
    JAWS 16
    NVDA 2016.1
    Any of Chrome, Firefox and IE 11

iOS 9.3.1 (Safari)
    iPhone 5s
    iPad mini 4

I don't have the means to test this with older versions of JAWS/IE etc. so there may be some limitations to this.

There are warnings against using this sort of technique at http://webaim.org/techniques/css/invisiblecontent/ though - it is possible older versions of assistive technology and browsers might not work. There is another technique in that article - css clipping, which I have seen used, but I don't know how well it is supported in older combinations.

robinwhittleton:
Thanks for testing this @accessiblewebuk. I’d forgotten about that WebAIM page; have updated my PR to include their 1px × 1px and `top: auto;` recommendations.

robinwhittleton:
@cfq did you ever get a chance to test this? Thanks!

cfq:
@robinwhittleton LGTM. Just tested it and tried to break it a little bit. Seems to work well. 
Thanks for implementing the WebAIM technique as well. 

👍

nickcolley:
Thanks everyone for comments and @robinwhittleton for the fix 👍  🎉

PR #178, opened 2015-09-16
Release 0.16.0

jamiecobbett:
/cc @froddd @edds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.16.0
+
+- Add Django template
+
 # 0.15.1
 
 - Fix copyright logo in Ruby version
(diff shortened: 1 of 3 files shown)


dsingleton:
Diff looks straight-forward: https://github.com/alphagov/govuk_template/compare/v0.15.1...release_0-16-0

I looks like `django` is missing from the [packaging](https://github.com/alphagov/govuk_template#packaging) section of the README, and i'm not sure if it'll appear in the [gh-pages](https://github.com/alphagov/govuk_template/blob/gh-pages/data/index.yml) doc either. If we're adding a new templating language those should probably be updated.

froddd:
Added Django to readme in https://github.com/alphagov/govuk_template/pull/179

dsingleton:
I think this is good to merge :+1: 

@froddd could you raise a PR on the `gh-pages` branch, to [add it there](https://github.com/alphagov/govuk_template/blob/gh-pages/data/index.yml#L48-L56) too? Thats the only remaining thing I think.

PR #183, opened 2015-11-26
Django package

froddd:
Add automatic publishing of the Django packaged version of the GOV.UK Template to a specific Github repository, so it can be used in Django projects. 

The Django packager has been updated to follow a more standard structure.

Main README updated to reference new package repository.

_**NOTE:** the new repository (alphagov/govuk_template_django) needs to be created by someone with write access for alphagov!_

**Todo:**
- [ ] Create alphagov/govuk_template_django Github repository
- [ ] Run publisher to update new repository


froddd:
Any movement on checking this? Could one of the admins review please? Thanks :)

quis:
Are you try to use pip to install the template? If so you don’t need a separate repo. This will have the same effect:
```shell
pip install https://github.com/alphagov/govuk_template/releases/download/v0.16.0/django_govuk_template-0.16.0.tgz
```

As stands I don’t think it will install because you don’t have a `setup.py` file, but you could generate one from [a template like this](https://github.com/quis/govuk_template/blob/add-python-package-files/source/setup.py.erb) with [a method like this](https://github.com/quis/govuk_template/blob/add-python-package-files/build_tools/packager/jinja_packager.rb#L28-L36).

quis:
(Something I forgot in my last comment) you need to [override the `create_tarball` method](https://github.com/quis/govuk_template/blob/add-python-package-files/build_tools/packager/jinja_packager.rb#L51-L58) as well, so that it extracts into the current directory, not a subdirectory.

Otherwise you get errors like this when trying to `pip install`:
```
No such file or directory: '/var/folders/1j/kwgs6fs16yz6s2c3h2mz67r80000gn/T/pip-MLMDnD-build/setup.py'
```

froddd:
@quis Any idea what that travis built fail might be about?

As you suggested I've now added the `setup.py` (I was convinced I'd added it before... oh well!) and overridden the `create_tarball` method as you mentioned. Tests pass nicely locally and installing the local package works with our app, too. 

The only reason I asked about creating the other repository was for consistency with the other publishers, which seem to do just that -- although as you mention it might be just as easy to point directly to the tarball -- I'm happy with that.

quis:
Checked out locally, works well, installs OK, tests pass.

The Travis issue looks to be a problem installing a Gem. Probably needs someone with more Ruby knowledge than me :eyes: @dsingleton…

dsingleton:
Looks like it's caused by an issue with the version of RubyGems on Travis,

I've got a [branch that fixes it](https://github.com/alphagov/govuk_template/commit/e7c8006978106d02b579047aec63ddf24a522670), but I want to make sure thats good way to fix it first. Hopefully have a PR today.

Once thats merged a rebase from master should fix the build here.

froddd:
@dsingleton I can see the fix has been applied (https://github.com/alphagov/govuk_template/pull/186), am I good to merge master into this and push again?

dsingleton:
> @dsingleton I can see the fix has been applied (#186), am I good to merge master into this and push again?

Yep!

quis:
Functionally this looks good to me.

Could someone with more Ruby knowledge run their eyes over the code?

benlovell:
Ruby bits look good to me. :ship:

fofr:
@quis Happy for this to be merged now?

quis:
:+1:

PR #185, opened 2016-01-13
[do not merge] Fix TravisCI builds by updating bundler

dsingleton:
The TravisCI version of bundler is out of date, which can cause some problems
install dependencies. Updating bundler as part of the install setup is the
fix recommended by Travis (See https://github.com/travis-ci/travis-ci/issues/3531)

I was seeing it another govuk_template branch:
- https://travis-ci.org/alphagov/govuk_template/builds/102056609

There is an identical fix for govspeak:
- https://github.com/alphagov/govspeak/commit/1135adab9d1e4404b9a903ee1ffb35f7ebbeb21e

More info:
- https://github.com/rubygems/rubygems/issues/1419
- https://github.com/thoughtbot/factory_girl/pull/846

PR #186, opened 2016-01-13
Move to Travis's container-based infrastructure

benilovj:
https://docs.travis-ci.com/user/workers/container-based-infrastructure/

This is an alternative to https://github.com/alphagov/govuk_template/pull/185

/cc @dsingleton

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+sudo: false
 install:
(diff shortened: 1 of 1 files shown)


dsingleton:
This seems like a nicer approach, no need for a pre-install tweak, and containerised builds should run faster. Good spot :+1: 

Happy to merge this once the builds are green and close mine.

Issue #188, opened 2016-01-25
Allow overriding / removal of OGL logo


andymantell:
I am working on a service for which I have been asked to remove the OGL logo and message from the footer. The reason being that some of the information contains personal information such as names and addresses and is not "open". In addition, users are required to purchase the information - it is not freely available.

The way the footer template is structured allows you to override the licence message, but not the OGL logo itself since it is outside the `if` statement:

```
<div class="open-government-licence">
  <p class="logo"><a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a></p>
  <% if content_for?(:licence_message) %>
    <%= yield :licence_message %>
  <% else %>
    <p>All content is available under the <a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence v3.0</a>, except where otherwise stated</p>
  <% end %>
</div>
```

Would it be reasonable to consider moving the logo inside the default message in the `else` block so that it can be removed / overriden? (Or some other way of disabling it). Or is the OGL logo intentionally baked in like this to discourage people from removing it?

fofr:
It looks like the custom licence message was part of a move to allow all strings to be customisable: https://github.com/alphagov/govuk_template/commit/dfcc0a34369f9a13905239fdcdc2ecc56504a568

The default text says, "except where otherwise stated”, a message to state what’s copyright/protected might work in conjunction with that?

Probably worth raising a PR to add a feature that allows the OGL logo to be removed, as it appears there's a clear need. Putting the logo within the `if` statement probably doesn't fix it – that would hide the logo for all custom licence messages, some of which could be OGL related.

andymantell:
Quite new to these gov uk templates. I was going to do something like this...

```
<% unless @omit_ogl_logo %>
  <p class="logo"><a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a></p>
<% end %>
```

...following the pattern of `<% unless @omit_header %>` on line 79 but of course many of the templating languages don't have this kind of logic built in. So, I'm wondering whether this might be more suitable?

```
<% if content_for?(:licence_logo) %>
  <%= yield :licence_logo %>
<% else %>
  <p class="logo"><a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a></p>
<% end %>
```

[...]

PR #189, opened 2016-01-26
Allow overriding / removal of OGL logo in the footer

andymantell:
As per #188, I am working on a service for which I have been asked to remove the OGL logo and message from the footer. The reason being that some of the information contains personal information such as names and addresses and is not "open". In addition, users are required to purchase the information - it is not freely available.

I have tried to follow the conventions used by the licence message in the footer but I am quite new to using the kit so not 100% sure if I have done this in the right way.

One thing I am aware of is that I might need to make a corresponding change to this file in the govuk_elements repo? https://github.com/alphagov/govuk_elements/blob/master/lib/template-config.js

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -110,7 +110,12 @@
             <%= yield :footer_support_links %>
 
             <div class="open-government-licence">
-              <p class="logo"><a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a></p>
+              <% if content_for?(:licence_logo) %>
+                <%= yield :licence_logo %>
+              <% else %>
+                <p class="logo"><a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a></p>
+              <% end %>
+
               <% if content_for?(:licence_message) %>
                 <%= yield :licence_message %>
               <% else %>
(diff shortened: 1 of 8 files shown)


source/views/layouts/govuk_template.html.erb
```diff
@@ -110,7 +110,12 @@
             <%= yield :footer_support_links %>
 
             <div class="open-government-licence">
-              <p class="logo"><a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a></p>
+              <% if content_for?(:licence_logo) %>
+                <%= yield :licence_logo %>
```
fofr: There doesn't seem to be an obvious way to remove the logo without providing an alternative.

How would you use this in your app?

andymantell: I guess I would override it with a zero length string in my local copy of this file:

https://github.com/alphagov/govuk_elements/blob/master/lib/template-config.js

Though I could be wrong. Part of the difficulty with this kind of thing is the need to target so many templating languages?! I.e. with mustache being "logicless"

fofr: Using an empty string doesn't feel like the right way of achieving this.

andymantell: @fofr I don't disagree with you, just not sure what the alternative would be within the constraints of the multiple templating languages we have to deal with?

Happy for this PR to be declined for now though until a better solution comes up...

nickcolley:
Hey @andymantell did you manage to work around this? Want to see if we can get this resolved..

andymantell:
@nickcolley Yeah I just edited the template we were using. Probably just worth declining this PR, it may only be me that even wanted this. And I don't know that my proprosed changes were necessarily the right way of doing it.

nickcolley:
Okay, feel free to reopen if anything changes 👍

PR #190, opened 2016-01-26
Add static files to `assets.precompile`

gpeng:
This avoids errors when gem is used with sprockets-rails >= 3

### DIFF
--- a/lib/govuk_template/engine.rb
+++ b/lib/govuk_template/engine.rb
@@ -2,11 +2,19 @@ module GovukTemplate
   class Engine < ::Rails::Engine
     initializer "govuk_template.assets.precompile" do |app|
       app.config.assets.precompile += %w(
+        favicon.ico
         govuk-template*.css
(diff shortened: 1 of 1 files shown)


lib/govuk_template/engine.rb
```diff
@@ -2,11 +2,20 @@ module GovukTemplate
...
         govuk-template*.css
         fonts*.css
         govuk-template.js
         ie.js
         vendor/goog/webfont-debug.js
+        apple-touch-icon-120x120.png
+        apple-touch-icon-152x152.png
+        apple-touch-icon-60x60.png
+        apple-touch-icon-76x76.png
+        favicon.ico
```
fofr: Looks like a duplication to line 5?

matthewford:
We also add to add the images under: https://github.com/alphagov/govuk_template/tree/master/source/assets/stylesheets/images

e.g. govuk-crest-ie.png

PR #193, opened 2016-02-01
Generate package.json for the Jinja version of the template

gemmaleigh:
This PR copies the package.json generator from the mustache packager.

Adding a `package.json` file ensures the [GOV.UK Prototype Kit](https://github.com/alphagov/govuk_prototype_kit/blob/master/package.json#L38) (amongst other projects) can use the Jinja version of the template as a dependency, without requiring [napa](https://www.npmjs.com/package/napa) as an additional dependency.

Napa is used a workaround, because the Jinja version of the GOV.UK template is missing the `package.json` file.

cc. @dsingleton @quis.

### DIFF
--- a/build_tools/packager/jinja_packager.rb
+++ b/build_tools/packager/jinja_packager.rb
@@ -20,6 +20,15 @@ module Packager
       end
     end
 
+    def generate_package_json
+      template_abbreviation = "jinja"
+      template_name = "Jinja"
+      contents = ERB.new(File.read(File.join(@repo_root, "source/package.json.erb"))).result(binding)
+      File.open(File.join(@target_dir, "package.json"), "w") do |f|
+        f.write contents
+      end
+    end
+
     def process_template(file)
       target_dir = @target_dir.join(File.dirname(file))
       target_dir.mkpath


quis:
I don’t know if there are potential problems with conflating Jinja and Nunchucks. The syntax isn’t identical, so this _could_ make changes more fragile in the future.

But, this _would_ be better for the Jinja projects I’m working on/have worked on. They depend on Bower, just for the sake of pulling in GOV.UK Template, and are already using NPM for everything else.

dsingleton:
:+1: to @quis's points.

This feels like a big improvement for programatically consuming the template, but conflating jinja/nunchucks. As I understand it this is adding `package.json` to the `jinja` package output, which prototype-kit will consume as if it's `nunchucks`, is that right?

If the jinja implementation did diverge from what nunchucks could support, then we could fork the packagers, having separates ones for jina/nunchucks that handle the differences, and consumers of them could update the package name, as well as version, when updating.

(Not 100% sure i've got that right, not super familiar with the actual consumers involved here)

quis:
@dsingleton You’ve got it right.

There’s a summary of the differences in the Nunchucks docs: http://mozilla.github.io/nunjucks/faq.html#can-i-use-the-same-templates-between-nunjucks-and-jinja2-what-are-the-differences

I can’t see any features there that we’d likely use, and half of them are things that are only relevant if the template was Nunchucks-first.

I’m going to do a quick dig into how easy integration tests would be for both, since neither have them at the moment.

quis:
Added a Jinja integration test: https://github.com/alphagov/govuk_template/pull/195

quis:
And one for Nunjucks: https://github.com/alphagov/govuk_template/pull/196

gemmaleigh:
:+1:

quis:
Checked out locally, rebased against `alphagov/master` and the tests pass. So :+1: from me as well.

image

gemmaleigh:
@dsingleton - the tests are passing - would :heart: to get this merged.

dsingleton:
Sorry, lost track of this one.

I think this is good to do, but spotted a small issue;

The [`package.json` template](https://github.com/gemmaleigh/govuk_template/blob/add-package-json-to-jinja-version/source/package.json.erb#L6-L11) assumes the there is an GH repo with `alphagov/govuk_template_< template_abbreviation>`, so in this case `alphagov/govuk_template_jinja` (which doesn't exist).

Existing packagers that user the `package.json` template (`ejs` and `mustache`) do have repos that do exist, but they also both published to the `npm` registry.

Having a 404 repo URL isn't the end of the world, and i'm not sure it should block this PR.

We could update the `package.json` template to make the repo URL conditional (not sure if thats valid), or make the `template_abbreviation` in the URL optional, and just point at the main repo.

quis:
:tada:

PR #194, opened 2016-02-02
Adds Support for Safari Pinned tabs

### LINKED ISSUE #192: Support for Safari Pinned tabs
[Safari now supports](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html) a new favicon type for pinned tabs, that uses SVG.

It might be helpful to add a favicon for this, for users who require a government resource on a regular basis and choose to pin the tab.

MikeRogers0:
This closes #192 by adding the SVG and required html snippet.

The SVG was cut out from https://en.wikipedia.org/wiki/File:Gov.uk_logo.svg as I couldn't find a similar SVG in the repo.

logo-demo


zaccolley:
looks nice! :fire:

fofr:
When does this SVG get downloaded? Do other browsers ignore the file?

I agree it looks nice, but is it necessary page weight for the feature it adds?

zaccolley:
Yet another apple icon with a weird offstandard meta tag aha.

Can see the support for other browsers here: https://github.com/whatwg/html/issues/110

I agree, if it's adding to the page weight it's not needed. Apparently there's a cache for the SVGs but still, favicons seem to suffice.

robinwhittleton:
As it’s got a `mask-icon` rel other browsers don’t seem to proactively download this resource (or at least Firefox and Chrome don’t). I’d be happy to see this go in, especially seeing as we already do touch icons.

PR #196, opened 2016-02-08
Add integration test for Nunjucks renderer

quis:
This commit adds an integration test for rendering the **Jinja** GOV.UK Template with the **[Nunjucks]( https://mozilla.github.io/nunjucks/)** template rendering engine. The test setup is based on the Mustache inheritance one.

The two templating languages are [mostly](http://mozilla.github.io/nunjucks/faq.html#can-i-use-the-same-templates-between-nunjucks-and-jinja2-what-are-the-differences) compatible. Having a test will ensure that we don’t get caught out by any of the incompatibilities.

This is to avoid having to maintain two separate, but identical versions of the template for the two different languages, as discussed in #193.

### DIFF
--- a/integration_tests/integrations/nunjucks/.gitignore
+++ b/integration_tests/integrations/nunjucks/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+vendor/*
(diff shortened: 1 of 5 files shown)


dsingleton:
:star2:

PR #201, opened 2016-02-16
Add new colour variables to replace deprecated greys

gemmaleigh:
Create a new set of Sass colour variables, which are used only by GOV.UK template.

Also delete the old `$grey` variables, which were overriding those set by the GOV.UK front end toolkit.

The `_base.scss` file used by GOV.UK elements has [been renamed and updated to only include base styles](https://github.com/alphagov/govuk_elements/pull/176) which aren't specific to the GOV.UK template - so it no longer copies the footer background colour on the html element.

### DIFF
--- a/source/assets/stylesheets/styleguide/_colours.scss
+++ b/source/assets/stylesheets/styleguide/_colours.scss
@@ -1,16 +1,12 @@
+// GOV.UK template colours
+
+$proposition-border: #2e3133;
 // Based on GOV.UK brand blue but slightly lighter so it hits contrast on the
 // black header bar when used as copy colour.
 $proposition-active-nav: #1d8feb;
 
-
-/* Old depricated greys, new things should use the toolkit greys */
-$grey-0: #0b0c0c;
-$grey-1: #171819;
-$grey-2: #2e3133;
-$grey-3: #454a4c;
-$grey-4: #5c6366;
-$grey-5: #8a9499;
-$grey-6: #a1acb2;
-$grey-7: #b8c6cc;
-$grey-8: #DEE0E2;
-$grey-9: #eaedef;
+$footer-background: $grey-3;
+$footer-border-top: #a1acb2;
+$footer-link: #454a4c;
+$footer-link-hover: #171819;
+$footer-text: $footer-link;
--- a/source/assets/stylesheets/_basic.scss
+++ b/source/assets/stylesheets/_basic.scss
@@ -45,7 +45,7 @@ html {
   -webkit-text-size-adjust: 100%; /* 1 */
   -ms-text-size-adjust: 100%; /* 1 */
   // Set background colour to match footer background
-  background-color: $grey-8;
+  background-color: $footer-background;
 }
 
 /*
--- a/source/assets/stylesheets/_footer.scss
+++ b/source/assets/stylesheets/_footer.scss
@@ -1,8 +1,8 @@
 /* Global footer */
 
 #footer {
-  background-color: $grey-8;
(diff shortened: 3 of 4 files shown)


source/assets/stylesheets/styleguide/_colours.scss
```diff
@@ -3,14 +3,14 @@
...
-$grey-4: #5c6366;
-$grey-5: #8a9499;
-$grey-6: #a1acb2;
-$grey-7: #b8c6cc;
-$grey-8: #DEE0E2;
-$grey-9: #eaedef;
+// Old depricated greys, new things should use the toolkit greys
+// $grey-0: #0b0c0c;  // $black, $text-colour
+$grey-1: #171819;     // Footer links - hover colour
+$grey-2: #2e3133;     // Borders under list items in proposition header
```
fofr: Why is this `grey-2` different to: https://github.com/alphagov/govuk_frontend_toolkit/blob/28a01aeb691c0935b6ffea3d248ddf2750b1eed5/stylesheets/_colours.scss#L87

For reference, from toolkit:
```sass
// Standard palette, greys
$black: #0b0c0c;
$grey-1: #6f777b;
$grey-2: #bfc1c3;
$grey-3: #dee0e2;
$grey-4: #f8f8f8;
$white: #fff;
```

fofr: Can we update to the toolkit greys as the comment suggests?

gemmaleigh:
@fofr, good question. I think we should ask @markhurrell.

gemmaleigh:
This PR leaves the history for context. It is useful to show where the $grey variables came from.

It can't be merged until there is a new release of govuk frontend toolkit (which includes $footer-background). I'll mark as do not merge for now.

dsingleton:
LGTM 👍

IMO it'd be better to squash all the commits into one, as the final diff is pretty small, and i'm not sure the extra commits add enough context to justify the noise.

robinwhittleton:
Yep, looks good to me. I guess we’ll also need to fix this in elements? https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_base.scss#L54

dsingleton:
> Yep, looks good to me. I guess we’ll also need to fix this in elements? https://github.com/alphagov/govuk_elements/blob/master/public/sass/elements/_base.scss#L54
>  Show all checks

Elements shouldn't need any of the footer specific colour variables. As I understand it, that file exists to duplicate the general-purpose/non-`govuk_template` specific SCSS (like accessibility helpers, etc) for projects that consume `govuk_elements`, but not `govuk_template` itself, in which case they shouldn't need a grey specific to the footer implementation of `govuk_template` - as they're not using it, if you get what I mean?

I'm going to try and put together a discussion PR to split the SCSS in `govuk_template` into 2 groups (but still within the repo).

1. Things specific to the design of `govuk_template` and it's markup, etc.
2. General purpose helpers that are useful to projects that don't use `govuk_template` and could be moved to FET in the future

robinwhittleton:
There’s [only one occurrence of `$grey-8` in govuk_elements](https://github.com/alphagov/govuk_elements/search?utf8=%E2%9C%93&q=grey-8) and it doesn’t seem to be defined anywhere.

Re your 2nd point, agreed.

dsingleton:
> There’s only one occurrence of $grey-8 in govuk_elements and it doesn’t seem to be defined anywhere.

Yep, it can be removed, it's usage there is a `govuk_template` "design feature", so can be dropped :+1:

PR #205, opened 2016-03-02
Remove media down mixin

gemmaleigh:
Delete the `_conditionals2.scss` partial and remove the `media-down` mixin.

The [media-down mixin is used in the _basic.scss partial](https://github.com/alphagov/govuk_template/search?utf8=%E2%9C%93&q=%40include+media-down%28mobile%29).

Update the _basic.scss partial to swap the order external link includes.

Tthe smaller external link style (16px) is defined first and is overridden at 641px and up
(tablet) by the larger size (19px).


robinwhittleton:
Excellent 👏

PR #208, opened 2016-04-19
Clean up and modernise govuk_template.html

fofr:
Use HTML5 preferred values.

Reduce page load by:
* removing HTML comments
* removing unnecessary script `type` attributes (https://www.w3.org/TR/html5/scripting-1.html#script)
* shortening charset declaration to HTML5 preferred value (https://www.w3.org/TR/html5/document-metadata.html#attr-meta-charset)

Also removes line breaks in the `<head>` tag and puts conditional comments onto a single line where appropriate.


tvararu:
👍

tombye:
👍

PR #212, opened 2016-05-15
Migrate viewport declaration from govuk_frontend_toolkit

robinwhittleton:
The viewport declarations for Opera and IE used to be defined in govuk_frontend_toolkit. It would be better if that was purely mixins and functions, and it makes sense that these declarations should live in the same place as the meta viewport declaration.

I looked into removing them completely, but there’s conflicting evidence as to what versions of Opera uses them, and at least IE10’s snap mode on desktop needs an @viewport.

### DIFF
--- a/source/assets/stylesheets/_basic.scss
+++ b/source/assets/stylesheets/_basic.scss
@@ -1,3 +1,11 @@
+@-ms-viewport {
+  width: device-width;
+}
+
+@-o-viewport {
+  width: device-width;
+}
+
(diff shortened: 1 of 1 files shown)


fofr:
Is it worth including something about the purpose of these viewport rules?

fofr:
Minor comment, otherwise :+1:

robinwhittleton:
That occurred to me last night. I’ll do that today.

robinwhittleton:
In the end it turns out that Opera doesn’t need this any more. Added a comment for the IE requirement.

PR #217, opened 2016-05-26
Use text-colour for headings in footer

fofr:
This used to be a custom grey variable set to `#0b0c0c` but was converted to `footer-text`, which made the headings too grey (see https://github.com/alphagov/govuk_template/pull/201)

Switch back to the darker colour to improve contrast, use $text-colour instead.

cc @gemmaleigh 

## Before
screen shot 2016-05-26 at 15 10 19

## After
screen shot 2016-05-26 at 15 10 47


nickcolley:
Looks good to me 👍

Issue #216, opened 2016-05-26
Remove debug webfont dependency


duncanhall:
The main `govuk_template.html` file has a dependency on `vendor/goog/webfont-debug.js`. This dependency is bundled with the included assets.

This file seems to be an old debug version of the Google webfont loader. A search for its usage suggests alphagov and its derivatives are the only users: [Search for 'google "webfont-debug.js"'](https://www.google.co.uk/search?q=google%20%22webfont-debug.js%22&rct=j).

While it currently serves it's purpose, there might be issues with using an old, debug version of the file. The [Google Web Font Loader project](https://github.com/typekit/webfontloader) suggests loading loading a production version using a CDN such as [googleapis.com](https://ajax.googleapis.com/ajax/libs/webfont/1.5.10/webfont.js), [jsDelivr](http://www.jsdelivr.com/projects/webfontloader) or [CDNJS](https://cdnjs.com/libraries/webfont). Alternatively, one of these newere, compressed versions could be downloaded and replace the current bundled version.

robinwhittleton:
I’ve done some spelunking, and (although I wasn’t at GDS when this happened), it looks like this file was [originally included](https://github.com/alphagov/static/commit/1898d4a79f3177fba236f0ae3fc22b0267b9b7aa) to help with loading Gill from Monotype. After [we stopped using Gill](https://github.com/alphagov/static/commit/93256fd246e2da5e6f525feaeb27af0db3021736) the google JS hung around for a while (eventually [for IE8 and lower](https://github.com/alphagov/static/commit/608bf1c0c6b77a8a73852e2a075162ffc62b21fe)), but when govuk_template was set up and gained fonts [they were loaded without JS](https://github.com/alphagov/govuk_template/commit/2b1ad04ba9855a832cac17dd23664092a7941172). The [loader was removed two months later](https://github.com/alphagov/static/commit/5297293ce11a58a172a4885b3b24845947df421c), but the library has remained part of govuk_template until now.

Conclusion: it should probably be removed.

PR #219, opened 2016-05-31
Remove JS font loader

robinwhittleton:
The JS font loader from Google appears to be vestigial. We can get rid of it, and rework the loading of fonts in IE8 to use just the CSS file. This appears to work for me in test environments, but if anyone has any more context as to why it’s stuck around then please comment before this gets merged.

Will fix https://github.com/alphagov/govuk_template/issues/216 . I did some investigation in https://github.com/alphagov/govuk_template/issues/216#issuecomment-222657594 which might be pertinent.


mcgoooo:
looks good to me :+1: although i would probs remove the ie6 include by now, surely :trollface:

accessiblewebuk:
I have learnt a new word "vestigial" - don't tell any content designers.

robinwhittleton:
Looking at the loader code, it sets (according to state) a set of `.wf-*` classes that let you style the site during the loading phase of the font. We’re not using any of these in the rest of the code. We’re also not using any of the provided callback hooks in our JS.

tvararu:
👌  Looks good to me.

nickcolley:
💣  👍

PR #221, opened 2016-06-16
Add logo svg asset to precompile list - fixes rails error.

dsingleton:
Using the latest release of the gem I was getting this error with
a brand new rails app:

```
Sprockets::Rails::Helper::AssetNotPrecompiled at /
Asset was not declared to be precompiled in production.
Add `Rails.application.config.assets.precompile += %w( gov.uk_logotype_crown.svg )` to `config/initializers/assets.rb`
```

The crown SVG was added relatively recently, but not included in the
rails asset precompile list, so i've added it.

Curiously this wasn't a problem with GOV.UK's `static` applicaton,
which has the same version of the gem, but runs fine.

Regardless, i think any assets referenced should be part of the
precompile list, so it's worth adding here.

### DIFF
--- a/lib/govuk_template/engine.rb
+++ b/lib/govuk_template/engine.rb
@@ -13,6 +13,7 @@ module GovukTemplate
         apple-touch-icon-76x76.png
         gov.uk_logotype_crown_invert.png
         gov.uk_logotype_crown_invert_trans.png
+        gov.uk_logotype_crown.svg
         opengraph-image.png
       )
     end


gemmaleigh:
I haven't fully tested this (by creating a new rails app) but it looks like [Verify are doing the same to fix this](https://github.com/alphagov/verify-frontend/blob/master/config/initializers/assets.rb#L11).

robinwhittleton:
If this goes in I guess I can remove it from Verify…

robinwhittleton:
Haven’t tested either, but I added the same fix on Verify to work around the same issue so happy to merge.

dsingleton:
Still unclear to me why `static` didn't have the same issue, but 👍

dsingleton:
For future reference, it depends on `sprockets-rails` version, which is different between `static` and `verify-frontend` (and my test app).

See https://github.com/alphagov/govuk_template/pull/190

PR #225, opened 2016-06-23
Add NPM publisher for jinja package

dsingleton:
New repo created at http://github.com/alphagov/govuk_template_jinja and initially updated by running this code locally.

Follows the same pattern as other NPM published versions, like `mustache` and `ejs`. There's quite a bit of duplication in the publisher code, but this doesn't feel like the time to refactor it, partly as we're looking at what might replace `govuk_template` as part of the Patters and Tools team discovery.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -146,6 +146,15 @@ namespace :build do
       puts "Pushing govuk_template_ejs #{GovukTemplate::VERSION} to git repo"
       q.publish
     end
+
+    require 'publisher/jinja_publisher'
+    q = Publisher::JinjaPublisher.new
+    if q.version_released?
+      puts "govuk_template_jinja #{GovukTemplate::VERSION} already released. Not pushing."
+    else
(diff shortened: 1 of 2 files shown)


gemmaleigh:
This looks to be the same as the other publisher files, for example, [the publisher for the mustache version of the template](https://github.com/alphagov/govuk_template/blob/master/build_tools/publisher/mustache_publisher.rb).

dsingleton:
Update, `npm` packages aren't being updated when we release a new version, due to some incorrect config in our CI server (which does the pushing).

The `npm` credentials were rotated recently, and although they were updated in [CI deployment config](https://github.gds/gds/ci-deployment/pull/80), this hasn't fully deployed to the CI, so the publish tasks are still using the old auth and failing.

GOVUK Infrastructure are looking into this at the moment, but once it's fixed we may need to do some cleanup to ensure any missed releases are published.

PR #227, opened 2016-06-23
Gem publisher

dsingleton:
This repo is natively a gem, and the other publishers are treated as derivatives of that. Long term we want `govuk_template` to be more language/package agnostic and built to support many build outputs, where `gem` is just one of them.

Treating the gem publishing in the same way as other types is a small step towards that. Also couples docs publication to the repo release step, rather than the gem.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,6 @@
 $:.unshift File.expand_path('../lib', __FILE__)
 $:.unshift File.expand_path('../build_tools', __FILE__)
 require "govuk_template/version"
-require "gem_publisher"
 
 desc "Compile template and assets from ./source into ./app"
 task :compile do
--- a/Rakefile
+++ b/Rakefile
@@ -95,20 +95,18 @@ namespace :build do
 
   desc "Build and release gem to gemfury if version has been updated"
   task :and_release_if_updated => :build do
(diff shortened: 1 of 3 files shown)


dsingleton:
We've noticed that the latest version of `govuk_template` (`0.18.1`) is missing from `rubygems`.

It looks like this change broke publishing, and that was the first release since then. 

The break is in the logic of the `and_release_if_updated ` where the release check for the [github release](https://github.com/alphagov/govuk_template/blob/master/build_tools/publisher/release_publisher.rb) was shared for the gem, as they both test the same thing to see if a release was already made - the presence of a tag on the repo. 

I split the github release apart from the gem release check for clarity, but it's causing [this line](https://github.com/alphagov/govuk_template/blob/master/Rakefile#L100) to be false, as the gh tag already exists from the github release.

I'm not sure if it's best to revert this entirely, or roll forward with a logic fix in `and_release_if_updated`. Splitting out the gem publisher still feels like a good thing, but introduced potential confusion, as it doesn't work quite the same way as other publishers. 

Thoughts?

CC @robinwhittleton, @gemmaleigh

dsingleton:
Going to revert this - we can always re-apply/cherry-pick from here later.

PR #233, opened 2016-08-08
Update README package links

dsingleton:
- Add NPM packge for jinja version, which was added recently
- Have a link to ruby gems for consistency


robinwhittleton:
Looks fine.

PR #235, opened 2016-08-09
Refactor README and split long documentation out

dsingleton:
Move long `README` sections to `docs/` directory and re-write the `README` a bit.

The `README` has grown a lot over time, making it harder to read. And as a
single page adding additional doc only adds to the problem.

Split specific doc sections out into their own directory, for clarity. This
is a convention used on GOV.UK projects and encouraged in the README
styleguide: https://github.com/alphagov/styleguides/blob/master/use-of-READMEs.md

I've done this now because I want to add more detailed documentation on the
publishing process, and document how to fix missing published packages, which
is likely to be super long.

Hopefully this will encourage more detailed documentation on the repo in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,155 +12,16 @@ You can [view a collection of auto-generated examples](http://alphagov.github.io
 
 The Ruby language (1.9.3+), the build tool [Rake](http://rake.rubyforge.org/) & the dependancy management tool [Bundler](http://bundler.io/)
 
-## Packaging
+## Detailed Docs
 
-At present this generates 9 output formats:
-
-1. a gem containing a Rails engine
-2. a tarball containing Play Framework templates
-3. a folder containing Mustache templates
-4. a tarball containing Liquid templates
-5. a tarball containing Mustache Inheritance templates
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,8 +1,11 @@
-# GovukTemplate
+# GOV.UK Template
 
 This provides a template containing the GOV.UK header and footer, and associated assets.
```
edwardhorsford: This might be outside of the scope of this PR, but it would be good for the main readme to contain something about *who* template is for. It's not clear (to me) that this is outward facing for services rather than just for GDS use.

 eg 'GOV.UK Template is meant for all services running under the service.gov.uk domain....'

nickcolley:
LGTM 👍

README.md
dsingleton: Absolutely agree. That, and quite a lot of other stuff, did feel out of scope though.

I'll steal/add your eg line though 👍

dsingleton:
Added @edwardhorsford's suggestion. Otherwise I think this is good to merge?

Issue #232, opened 2016-07-14
Logo breaks layout with increased font size


robinwhittleton:
WCAG2 1.4.4 says that to hit double-AA [text must be able to zoom to 200% of the default size without breaking the site layout](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html). Currently the logo is fixed to 30px which breaks the text when resized:

![screen shot 2016-07-14 at 15 27 35](https://cloud.githubusercontent.com/assets/7414/16843029/87665130-49d7-11e6-8d54-a5f75d2d5514.png)

We should allow the header to grow in size to contain the text, probably by using `min-height` instead of `height`. Reported by @selfthinker.

accessiblewebuk:
This isn't a WCAG 2 failure in itself, because the text isn't obscured and the contrast between the white text and the blue background is still sufficient. Obviously it would be best practice for the layout not to break.

nickcolley:
Fixed by #237 thanks @selfthinker 🎉

PR #237, opened 2016-08-11
Logo fixes

selfthinker:
I initially only wanted to fix that the logo breaks layout with increased font size (#232).
But because I found a couple of small issue around the logo (invalid, never executed, non-standard or confusing code), I fixed and refactored a few more things. One of the changes includes changing the dimensions of the logo. The "new" logo (both versions) were done and provided to me by @markhurrell.

Three known issues with the new logo:
* gov.uk_logotype_crown.png was not transparent before but is now. I don't know if that is intentional. I will ask Mark when he's back.
* <del>I have intentionally not changed the IE8-and-under version of the logo as it doesn't look broken. Please let me know if you think that is important to do.</del>
* The only reason why the logo needs some fiddly pixel-pushing is because the logo itself includes whitespace. I will ask Mark when he's back if we could change that. (I would probably better create a new PR for that case.)

I looked at the git history for most of my changes, please refer to the commit messages for further information (history and rationale).

## Before/after screenshots

These screenshots were all made from Firefox on Mac.
The visible change is that the logo is grew 1px on both sides. Its bottom is still vertically aligned with the "G", so that it grows 1px to the top and 1px to the right. But the top space was removed again from an containing div, so the height dimensions of the left-hand elements should be exactly the same as before (except for the image itself), the width got pushed 1px to the right.

Screen logo before:
logo-screen-before2

Screen logo after (hardly any changes, only the logo is bigger):
logo-screen-after3

Print logo before:
logo-print-before2

Print logo after (a few more changes, for the better I think):
logo-print-after2

## Browser tests

I tested in a lot of browsers via BrowerStack and it generally was fine. Some probably ignorable issues I found with this change:

* iPhone 4 was not perfectly aligned, but iPhone 4S is
* Kindle was not perfectly aligned either but that has other general issues as well
* Galaxy S5 Mini (4.4.) dropped the "GOV.UK" below the logo, but Galaxy S4 (4.4) was fine again

What I didn't test: Because I don't know what else could potentially affect the header (other included elements or certain modifications) I didn't test any of that. It would be great if someone could let me know what else there is and how I can test it.


source/assets/stylesheets/_core-print.scss
```diff
@@ -37,23 +37,18 @@ img { max-width: 100% !important; }
 select { background: white; }
 
 #global-header {
-  color: #000;
-  font-weight: bold;
-  font-size: 24px;
-  margin-bottom: 1em;
+  margin: 3px 0 1em; // 3px addtional space at top for countering -3px re-poositioning of #logo img
```
robinwhittleton: ‘addtional’ should be ‘additional’. ‘re-poositioning’ should be ‘re-positioning’

source/assets/stylesheets/_header.scss
```diff
@@ -71,44 +71,26 @@
...
-    margin-bottom: -3px;
-    padding-top: 2px;
     border-bottom: 1px solid transparent;
 
     background: image-url('images/gov.uk_logotype_crown.png') no-repeat;
-    background-size: 35px 31px;
+    background-size: 36px 32px;
     background-position: 0 0;
     @include ie-lte(8) {
       background-image: image-url('images/gov.uk_logotype_crown-1x.png');
```
robinwhittleton: No update to the IE8 crown image?

robinwhittleton:
Is there a reason why `gov.uk_logotype_crown.png` is using 1-bit transparency instead of translucent edges?

robinwhittleton:
Also, did we consider moving to an SVG crown? Would be good to document why we decided not to here.

selfthinker:
Regarding the transparency, I don't know. I can use graphic tools to a degree but felt not confident enough doing something with an image as important as the logo. That's why I asked Mark to make them. Let's ask him about the transparency when he's back.

We talked about using an SVG but if we wanted to do that, I think that should be in a separate PR. All of these commits here are fixing one thing or another. Using an SVG logo would be more of a feature than a bugfix.

source/assets/stylesheets/_core-print.scss
selfthinker: Oops, fixed.

markhurrell:
hey both, used the 1bit transparency out of habit (was an old IE thing right?). it's on a black background so thought might as well put it on a black matte. happy to do it again with full transparency if you prefer.

being nit-picking, the GOV.UK looks like its sitting a tiny bit low on the new version (I guess this is because there's more whitespace around the top of the search bar now). could we shift it up a pixel or two?

robinwhittleton:
IE6 and lower. I think we can do full translucency now :)

selfthinker:
@markhurrell, sorry, I think I might have missed a pixel or so from the top when making the screenshot. Because nothing has changed on the dimensions at all, except what I described earlier:
> The visible change is that the logo is grew 1px on both sides. Its bottom is still vertically aligned with the "G", so that it grows 1px to the top and 1px to the right. But the top space was removed again from an containing div, so the height dimensions of the left-hand elements should be exactly the same as before (except for the image itself), the width got pushed 1px to the right.

I validated that by making a screenshot (together with bits of the chrome) and measuring the pixels while zoomed in.

I guess it's probably best if we could look at it together on a real screen (and not a screenshot) to be sure. I can also push some pixels around at the same time if I need to. That should be the quickest to resolve any potential issues.

markhurrell:
govuk-crown

[full transparency PNG attached above]

selfthinker:
@markhurrell, are you sure that is the correct image? I just put it into the folder and my git doesn't show me that there is any difference.

On another related note, I have made a new before/after gif which shows both versions after each other, together with the browser canvas on the side to make sure you can see the correct dimensions. (Sorry for the bad quality.)
logo-before-after

alextea:
@robinwhittleton the 1-bit transparency could be a side effect of using ImageAlpha incorrectly.

markhurrell:
nope I did it deliberately because I care about ie6 users bruh

--

*Mark Hurrell*
Head of Design for GOV.UK
Government Digital Service

selfthinker:
I have updated the logo with the new one Mark provided. (Sorry, I didn't see any changes in git because I failed to copy the file correctly into the corresponding folder.) I also updated the one for IE8 and under (I just halved the new one).

nickcolley:
Looks good, only thing I've noticed is that on the print styles in Chrome the logo is not as expected.

Before

screen shot 2016-09-19 at 17 34 43

After

screen shot 2016-09-19 at 17 33 47

selfthinker:
@nickcolley, I believe this should fix that issue: alphagov/govuk_frontend_toolkit/pull/328

selfthinker:
The other PR was merged, but I'm not sure how long it takes for the frontend toolkit to be packaged and released.
@nickcolley, now that that last issue is fixed, would you rather a) wait with the merge of this PR until the other change is live, or b) merge this now anyway, or c) undo the offending two lines and then merge this and make the print change in another PR which can wait until the toolkit has been released?

nickcolley:
@selfthinker we should wait till upstream is up to date then update it here to fix the issue I think?

selfthinker:
@nickcolley, as we investigated and found that [govuk_template uses a **16 months old** frontend toolkit](https://github.com/alphagov/govuk_template/blob/master/govuk_template.gemspec#L24) and I don't think it's worth the hassle, I have removed the offending line and left a comment in the CSS accordingly.

I hope this can now be merged?

nickcolley:
LGTM 👍

PR #242, opened 2016-08-25
Remove color within link attribute

surminus:
In accessability testing it was flagged that a color attribute was within a <link> attribute which is against the W3 HTML guidelines: https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.gov.uk%2F

This was added in commit 0d300da1bdc470527a35a3501037672974741505 along with other changes but without explanation about why the color part was added specifically.

Removing this on an HTML page doesn't *seem* to make any significant changes to the template.

Note: haven't been able to test this locally on dev VM so would be interested to see how this is tested, if required.

/cc @emmabeynon @klssmith @davidbasalla

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -17,7 +17,7 @@
     <!--[if lt IE 9]><script src="<%= asset_path "ie.js" %>"></script><![endif]-->
 
     <link rel="shortcut icon" href="<%= asset_path 'favicon.ico' %>" type="image/x-icon" />
-    <link rel="mask-icon" href="<%= asset_path 'gov.uk_logotype_crown.svg' %>" color="#0b0c0c">
+    <link rel="mask-icon" href="<%= asset_path 'gov.uk_logotype_crown.svg' %>" >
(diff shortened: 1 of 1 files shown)


nickcolley:
Looks like the attribute provides the fill for a icon in Safari pinned tabs.

https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html

I believe the only real fix here is to remove the entire meta tag entirely.

robinwhittleton:
This is non-standard, but it’s a non-standard feature. It’s specifically there to provide support for [pinned tab icons in Safari](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html). We don’t want to remove this, even though it warns on validation – it’s not an accessibility issue.

nickcolley:
For a look to the future, hopefully we'll have Web Manifests so we don't need to litter our markup with meta tags related to icons :)

https://developers.google.com/web/updates/2014/11/Support-for-installable-web-apps-with-webapp-manifest-in-chrome-38-for-Android?hl=en

PR #246, opened 2016-08-30
Add basic guide to using govuk_template with Rail

dsingleton:
This is based on a recent pairing/research session with a new developer
setting up `govuk_template` for the first time.

It doesn't aim to be 100% comprehenssive, but to cover the basics. It's
something we can research and build upon, hopefully

### DIFF
--- a/docs/using-with-rails.md
+++ b/docs/using-with-rails.md
@@ -1,0 +1,40 @@
+# Using with Rails
+
+## Add to Gemfile
+
+Add `govuk_template` to your app's `Gemfile`:
+
+```ruby
+gem 'govuk_template'
+```
+
+(In a production system you'll want to pin a more specific version for stability)
+
+## Use the layout
+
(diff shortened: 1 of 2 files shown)


docs/using-with-rails.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+(In a production system you'll want to pin a more specific version for stability)
+
+## Use the layout
+
+Add this line to the bottom of your application layout view (usually in `app/views/layouts/application.html.erb`):
+
+```erb
+<%= render :file => 'layouts/govuk_template' %>
```
robinwhittleton: Old-style hashrockets or new-style `<%= render file: 'layouts/govuk_template' %>`?

robinwhittleton:
Looks good, minus one nit. Will merge if you fix that (or tell me you don’t want to 😄)

dsingleton:
> Looks good, minus one nit. Will merge if you fix that (or tell me you don’t want to 😄)

Fixed - good catch.

PR #247, opened 2016-08-31
Updates to the template blocks documentation

gemmaleigh:
- Fix top_of_page typo
- Where there is an insertion point, which has default content -
explain what that content is, rather than where it occurs in the
document
- Add a note about the main content requiring an ID of #content for the
skiplink to work properly

### DIFF
--- a/docs/template-blocks.md
+++ b/docs/template-blocks.md
@@ -17,7 +17,7 @@
 | inside_header             | Inside parent `.header-global`                  | Insertion point
 | proposition_header        | Inside parent `.header-wrapper`                 | Add a [propositional title and navigation links](docs/usage.md#propositional-title-and-navigation)
 | after_header              | After closing `</header>` element               | Insertion point
-| content                   | Main content goes in here                       | Insertion point. Note: A wrapper with an ID of #content is needed here for the skiplink to work correctly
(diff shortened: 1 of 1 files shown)


docs/template-blocks.md
```diff
@@ -2,24 +2,24 @@
...
 | homepage_url              | URL of anchor element wrapping logo             | https://www.gov.uk/
 | logo_link_title           | Title of anchor element wrapping logo           | Go to the GOV.UK homepage
 | global_header_text        | Text next to the crown image                    | GOV.UK
 | inside_header             | Inside parent `.header-global`                  | Insertion point
-| proposition_header        | Inside parent `.header-wrapper`                 | Insertion point
+| proposition_header        | Inside parent `.header-wrapper`                 | Add a [propositional title and navigation links](docs/usage.md#propositional-title-and-navigation)
 | after_header              | After closing `</header>` element               | Insertion point
-| content                   | After `#global-header-bar`                      | Insertion point
+| content                   | Main content goes in here                       | Insertion point. Note: A wrapper with an ID of #content is needed here for the skiplink to work correctly
```
dsingleton: There's a usage guide on skip links we could cross-link: https://github.com/alphagov/govuk_template/blob/master/docs/usage.md#skip-link

PR #249, opened 2016-09-12
Version 0.18.2

robinwhittleton:
- Degrade gracefully when external JS can’t be loaded ([PR #248](https://github.com/alphagov/govuk_template/pull/248))


gemmaleigh:
👍

PR #250, opened 2016-09-15
Add docs for adding tabindex="-1" to fix the skiplink

gemmaleigh:
The GOV.UK template sets a skiplink to `#content` (just tab on any
GOV.UK page and hit ENTER when "Skip to main content" is highlighted).

This allows users to navigate to the main content area using only the
keyboard. Pressing enter when focus is on the skiplink should then set
focus on the #content area.

In Safari, there's a bug where this doesn't happen unless
`tabindex="-1"` is set on the destination element (in this case `<main
id="content">`). 

Here's the [bug report](https://bugs.webkit.org/show_bug.cgi?id=141113).

This pull request adds documentation for adding `tabindex="-1"` 
to the `<main>` element to fix this issue.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -8,6 +8,13 @@ Refer to the table of [template blocks](./docs/template-blocks.md) and their def
 
 The [govuk_template sets a skip link](https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L64-L68) to `#content`, but doesn't provide an element with `id="content"`. You'll need to add `id="content"` to your main content area, to ensure the skip link will work.
 
+    <main id="content" role="main" tabindex="-1">
+
+It is recommended to use the main element, with a role of main (to support older assistive technology), also `tabindex="-1"` is required to fix the issue [with focus in Safari](https://bugs.chromium.org/p/chromium/issues/detail?id=37721).
+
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -8,6 +8,13 @@ Refer to the table of [template blocks](./docs/template-blocks.md) and their def
 
 The [govuk_template sets a skip link](https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L64-L68) to `#content`, but doesn't provide an element with `id="content"`. You'll need to add `id="content"` to your main content area, to ensure the skip link will work.
 
+    <main id="content" role="main" tabindex="-1">
```
nickcolley: Do we need the tag to be `main` if we've got role="main" applied? Just thinking `main` has to be shimmed in older browsers..

gemmaleigh: We provide the [HTML5Shiv bundled with the govuk_template JS](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/vendor/html5shiv.js) and have added `display:block`for main to the [govuk_template css](https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_basic.scss#L16).

nickcolley:
Thanks @gemmaleigh 👍

PR #254, opened 2016-10-10
Don’t include both html5shiv and html5shiv-printshiv

robinwhittleton:
Printshiv already includes html5shiv, so including both duplicates the amount of code IE has to downloaded, parsed and run.


nickcolley:
https://github.com/aFarkas/html5shiv#html5shiv-printshivjs

```
This includes all of the above, as well as a mechanism allowing HTML5 elements to be styled and contain children while being printed in IE 6-8.
```

Nice, less JavaScript run on older browsers the better! 👍

PR #259, opened 2016-10-26
Have focus outline appear outside of element rather than covering it.

edwardhorsford:
This PR adds a 0 offset to focus outlines. This works around a bug in Chrome and Safari where focus is drawn over an element rather than around it.

Before:
screen shot 2016-10-26 at 18 00 03
screen shot 2016-10-26 at 18 00 27

After:
screen shot 2016-10-26 at 18 00 10
screen shot 2016-10-26 at 18 00 36

I've not tested it directly (don't know how to run template), but have made the same changes in browser to test. Does not seem to impact on buttons.

### DIFF
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -91,13 +91,19 @@ a:focus {
 input:focus,
 textarea:focus,
 select:focus,
-button:focus,
-#global-header input:focus {
+button:focus {
   outline: 3px solid $focus-colour;
   outline-offset: 0;
 }
 
 #global-header {
+
+  input[type=search]:focus {
+    outline: 3px solid $focus-colour;
+    /* Focus appears inside the input */
+    outline-offset: -3px;
+  }
+
   h1 a:focus {
     background-color: transparent;
     outline: none;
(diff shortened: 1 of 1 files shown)


nickcolley:
Wonder if this'd make this more consistent: https://github.com/alphagov/govuk_elements/issues/324

gemmaleigh:
You can preview these changes here:

http://govuk-elements-wip.herokuapp.com/form-elements/

http://govuk-elements-wip.herokuapp.com/form-elements/example-form-elements/

source/assets/stylesheets/_accessibility.scss
```diff
@@ -91,12 +91,19 @@ a:focus {
...
   outline: 3px solid $focus-colour;
+  outline-offset: 0;
 }
 
 #global-header {
+
+  input[type=search]:focus {
+    outline: 3px solid $focus-colour;
+    /* Focus appears inside the input */
+    outline-offset: -3px;
```
robinwhittleton: Making this `-3px` exposes a white line to the right. It’s technically correct, but given that this would require the refactoring of where search boxes live and how they’re styled across multiple services can we make it `-2px` for this PR? That normalises it across browsers in my (quick) testing. If you want to leave a comment do describing why that’s fine.

PR #261, opened 2016-11-10
Fixed scala compilation failure for play template

akhettar:
I have changed the followings:

- Made some of the parameters optional as the template expect them to be.

- Using double quotation marks when calling cookieMessage.getOrElse.

I have compiled the template using the following command: "bundle exec rake build:play" and tested it on prototype. All works fine now


robinwhittleton:
Thanks for this! I’m trying to track down someone else who uses Scala (we don’t internally at GDS) to test this. If I can’t then I’ll see if I can set something up to test.

Can you let me know what the original issues you’re fixing were?

robinwhittleton:
Happy to merge with @philsherry’s OK. Thanks for the contribution!

PR #263, opened 2016-12-05
Skiplink colour contrast

cfq:
Skiplink has the default link colour, which kept failing against $focus-colour
background. Making the link colour $black for :focus and :visited states.

### DIFF
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -65,6 +65,12 @@ a:focus {
 .skiplink {
   position: absolute;
   left: -9999em;
+
+  /* Default link colour doesn't have enough contrast against $focus-colour */
+  &:focus, 
+  &:visited {
+    color: $text-colour;
+  }
 }
 
 .skiplink:focus {


nickcolley:
So this is a problem for all links when focused. This seems like it could be fixed globally?

edwardhorsford:
@nickcolley for the skiplink, the text is already on a black background, so having it be black works well. I don't know if the same solution would work for the other cases.

source/assets/stylesheets/_accessibility.scss
```diff
@@ -65,6 +65,11 @@ a:focus {
 .skiplink {
   position: absolute;
   left: -9999em;
+
+  /* Default link colour doesn't have enough contrast against $focus-colour */
+  &:focus, &:visited {
+    color: $black;
```
edwardhorsford: This should probably be $text-colour (even though it maps to the same hex colour).

cfq:
@nickcolley @edwardhorsford: Just had a chat with @accessiblewebuk - he says for ordinary links, he wouldn't fail them for colour contrast if the contrast is only low on focus states, but it's true that the focus states don't have high enough contrast.

Would that be more sensible to add this to a:focus rather than just the skiplink? So the links would be $text-colour when focused.

source/assets/stylesheets/_accessibility.scss
```diff
@@ -65,6 +65,11 @@ a:focus {
 .skiplink {
   position: absolute;
   left: -9999em;
+
+  /* Default link colour doesn't have enough contrast against $focus-colour */
+  &:focus, &:visited {
```
robinwhittleton: Can you split the selectors onto separate lines to match our linting rules?

PR #267, opened 2016-12-23
Switch from using GOV.UK Jenkins to Travis

alexmuller:
This project isn't owned by GOV.UK Publishing anymore so it needs to stop using Jenkins to publish changes.

Ideally I'd use a deploy key for this, but it turns out that you can't attach the same deploy key to more than one repo. This means that we're going to use a GitHub token attached to the [govuk-ci](https://github.com/govuk-ci) user - this should be moved to a different user in the future. To make that happen:

- Add the personal access token as a secure Travis env var
- Change the Git URLs to push over HTTPS, not SSH

The gem_publisher gem expects there to be a file at `.gem/credentials`. I've encrypted that file for Travis and added that file to the Travis worker.

Same deal for the npm publish step.


alexmuller:
I've deleted the webhook and disabled the Jenkins job.

alexmuller:
Yay, it's working now I think. Had some trouble with encrypting multiple files for a while.

alexmuller:
Add "DO NOT MERGE" so that this stops notifying my team's Slack until somebody is ready to review it.

.travis.yml
```diff
@@ -1,11 +1,27 @@
 language: ruby
 sudo: false
+env:
+  global:
+    - secure: "gxjwgl9l9ZP/1t86odZyMzWyiVBkG1oFzeL+Kb6z9Dtfu954l9RbytlenTTcnRw1KtaDBZVBs0ix3K45b8T2S6JFzLQ88Z/36DmXGIN6vxctfnVD9sqbbyOuP++RhibK+qUqbPRHo6a3wBsMQ87DMj5S6pJe3dxbCKRXa8zcMd0="
+before_install:
+  - openssl aes-256-cbc -K $encrypted_df4ab1bff570_key -iv $encrypted_df4ab1bff570_iv -in .travis/secrets.tar.enc -out .travis/secrets.tar -d
```
dsingleton: This will need to move the deploy steps from `before_install` to `before_deploy`, as ENV vars won't be available when running CI for forked PRs, which are common on this repo.

See https://github.com/alphagov/govuk_frontend_toolkit/pull/385

.travis.yml
```diff
@@ -1,11 +1,27 @@
...
+  - tar xvf .travis/secrets.tar
+  - mkdir -p ~/.gem/
+  - mv secrets/gem_credentials ~/.gem/credentials
+  - mv secrets/npmrc ~/.npmrc
```
dsingleton: I can't see the decrypted version of `npmrc`, but I suspect it might have the same npm v1 vs v2 issue as in https://github.com/alphagov/govuk_frontend_toolkit_npm/pull/16#issuecomment-283359364.

.travis.yml
dsingleton: Turns out `before_deploy` is run _per deploy provider_, and running this setup twice will cause an error, as we discovered on `govuk_elements`. 

The most reliable way to limit decryption to situations where deployment is required is using `if [ "$TRAVIS_PULL_REQUEST_BRANCH" = "master" ];`, but this may not work well for multi-liners. See https://github.com/alphagov/govuk_elements/pull/415.

dsingleton:
> This means that we're going to use a GitHub token attached to the govuk-ci user - this should be moved to a different user in the future

Should move from `govuk-ci` to a different user? That seems like a good fit for now.

Otherwise, i think we're happy with the changes and the next step is to merge and test the various deployment setup/providers - as we have with other repos.

PR #268, opened 2017-01-10
Add placeholder to inline script tags for `nonce` value

JoeChapman:
Sign inline script tags with `nonce` to meet server's potential CSP directives

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -30,7 +30,7 @@
   </head>
 
   <body<%= content_for?(:body_classes) ? raw(" class=\"#{yield(:body_classes)}\"") : '' %>>
-    <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
+    <script<% if @nonce %> nonce="<%= nonce %>"<% end %>>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
 
     <%= yield :body_start %>
 
@@ -103,6 +103,6 @@
     <%= yield :body_end %>
 
     <%# if no GOVUK-namespaced module has loaded we can assume JS has failed and remove the class %>
-    <script>if (typeof window.GOVUK === 'undefined') document.body.className = document.body.className.replace('js-enabled', '');</script>
+    <script<% if @nonce %> nonce="<%= nonce %>"<% end %>>if (typeof window.GOVUK === 'undefined') document.body.className = document.body.className.replace('js-enabled', '');</script>
   </body>
 </html>


davbo:
Hi @JoeChapman it's great to see you contributing to improve `govuk_template`'s CSP support.

When we deployed [CSP on Verify](https://github.com/alphagov/verify-frontend/blob/master/config/application.rb#L38-L52) we used hashes as recommended by @robinwhittleton as these rarely change. It would be good to hear why this wouldn't work for you?

CSP is moving quite quickly, on one side the [spec](https://www.w3.org/TR/CSP/#strict-dynamic-usage) and [Google](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/45542.pdf) seem to be suggesting nonce-based policies are best as they allow for easier adoption of CSP (which I generally agree with) but at the same time people are finding many ways of [bypassing nonce-based policies](http://sebastian-lekies.de/csp/bypasses.php). 

Again I'm not opposed to adding nonce support but would like to know why the hashes wouldn't work for you before we do that.

Thanks! 

ps. moved comment to the correct place!

JoeChapman:
Hi @davbo and thanks for reviewing this issue/PR.

From our perspective, nonces are easy to maintain. Once we've added the nonce-generating middleware to the server - which sets a new nonce on the locals objects with each page load - and  added a nonce placeholder attribute to each script tag, future maintenance involves adding a nonce placeholder attribute to all new script tags.

There would be no requirement to make new releases when the nonce changes (it changes each page load). There wouldn't be a list of hashes to maintain.

This is an interesting document http://sebastian-lekies.de/csp/bypasses.php, can hashes also be bypassed?

davbo:
I'll emphasize that the scripts in `govuk_template` aren't likely to change often so maintaining the list in this case is fairly easy. I would like to make a machine readable version of the hashes ship in the packages though.

Hashes have a reduced attack surface (so aren't as likely to be bypassed) since they identify a single <script> block, whereas if you're able to access the nonce, which is possible in many ways including [exfiltrating it via CSS](http://sirdarckcat.blogspot.co.uk/2016/12/how-to-bypass-csp-nonces-with-dom-xss.html) then you're able to whitelist any script bypassing CSP.

By the way it looks like the changes here have broken the Django templates, if you can get it working I think we'd be happy to accept the change. I'd hate to deter your efforts here, I'm just hoping to explain some of the drawbacks of using the nonces and provide some examples (verify) of alternatives.

Thanks!

JoeChapman:
Thanks @davbo, I appreciate your help. 

> I'll emphasize that the scripts in govuk_template aren't likely to change often so maintaining the list in this case is fairly easy
There is a chance templates downstream may change to include more inline scripts.

> By the way it looks like the changes here have broken the Django templates
Yep, I have some debugging to do

Thanks again

sirdarckcat:
Hi

Sorry to intrude in the conversation. I wrote some of the bypasses mentioned above, and can confirm most of our attacks against CSP nonces don't work against hashes.

That said, the 'strict-dynamic' variants of the attacks should work against hashes too.

davbo:
Hi @JoeChapman,

Could you update your changes to keep the body of the script tag on a single line.

I know it'll make it a bit ugly but we run the risk of breaking the aforementioned rarely changing hashes if you add newlines to the body of the script (the hashed content is the `text` attribute, which will have newlines/spaces added).

Sorry I should have spotted that previously.

Thanks for fixing the breakage in Django templates!

davbo:
Hi @sirdarckcat,

Your input is definitely valued, thank you for commenting!

JoeChapman:
Hi @davbo thanks for your help getting this done. I've made the lone change you asked for.

JoeChapman:
@davbo apologies, and fixed.

robinwhittleton:
Just tested your branch with building some of the other packages, and the `@nonce` declaration isn’t surviving the transpilation. Is there a reason we’re not using the same method as with the body classes on the line before? For example:
```erb
<script<%= content_for?(:nonce) ? raw(" nonce=\"#{yield(:nonce)}\"") : '' %>>
```
That would let you just yield the nonces as part of your layout initialisation.

JoeChapman:
@robinwhittleton using a block declaration would cause the compilation step to fail because the nonce is declared more than once, and could potentially be declared an unlimited number of times.

JoeChapman:
@davbo are you happy with this?

JoeChapman:
@davbo any changes required?

davbo:
Hi @JoeChapman sorry for not getting back to you sooner.

I don't fully understand the way the transpilation @robinwhittleton mentioned works in this repo. If as Robin commented the nonce doesn't survive transpilation then is the change useful for yourself / others?

robinwhittleton:
Honestly, while I like the idea, I’m not particularly happy about making functionality available in some templating languages and not others. In my example could you just have different names for different nonce sections? Or am I missing something here?

JoeChapman:
Hi @robinwhittleton how would it work with hashes, because I suspect that the attribute name for each has should be the same, like a class?

timpaul:
Thank you for this work.

However, following the launch of the [GOV.UK Design System](https://design-system.service.gov.uk/), GOV.UK Template will now only get major bug fixes and security patches, so I'm going to close this PR.

PR #272, opened 2017-01-30
Improve contrast of links when focused

### LINKED ISSUE #264: Focused links have low contrast
Links don't have good contrast against $focus-colour when they're in focus, it's currently 4.07:1. Keyboard users who have low vision would notice them. Screen magnification and keyboard users whose zoom follow keyboard focus would find it hard if they need high contrast.

We can fix it by setting the link colour on focus to be the same as $text-colour.

Current:
screen shot 2016-12-07 at 15 27 17

Proposed:
screen shot 2016-12-07 at 15 26 11

edwardhorsford:
Fixes #264 

- Makes link text slightly darker when focused - now hits 1:4.5 contrast against yellow.

Note - I've not tested this with template itself, but the sass itself seems to work.

## Before:
screen shot 2017-01-30 at 13 08 42-normal

## After:
screen shot 2017-01-30 at 13 09 03-darker

### DIFF
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -60,6 +60,11 @@ a:focus {
   outline: 3px solid $focus-colour;
 }
 
+/* Make links slightly darker when focused to improve contrast. */
+a:link:focus {
+  color: darken( $link-colour, 2.5%)
+}
+
 /* Make skiplinks visible when they are tabbed to */
 
 .skiplink {


source/assets/stylesheets/_accessibility.scss
```diff
@@ -60,6 +60,11 @@ a:focus {
   outline: 3px solid $focus-colour;
 }
 
+/* Make links slightly darker when focused to improve contrast. */
+a:link:focus {
```
selfthinker: What should be the colour when the link is visited? If there is nothing overwriting that, visited links will probably still show the lighter blue.

edwardhorsford: I'm not proposing changing the visited colour when focused. It will stay as purple - which meets our contrast requirements.

robinwhittleton:
I’m happy with the first commit in this PR. The second one seems to be a dupe of https://github.com/alphagov/govuk_template/pull/274? This should either be revised to remove the second commit, or the other PR closed (and the suggestions in it applied here).

gemmaleigh:
@robinwhittleton @edwardhorsford I've amended this PR, so it now only makes the one change - to make link text slightly darker when focused.

edwardhorsford:
I've updated the text of the PR to reduce the scope of what it covers.

PR #274, opened 2017-01-30
Make header text colour black when focused

edwardhorsford:
The header text has poor contrast when focused as it's white on yellow. This PR changes it to black on yellow to match other anchors on the page.

Also changes to use `$text-colour` from black, which felt more consistent.

Before:
screen shot 2017-01-30 at 13 04 42
After:
screen shot 2017-01-30 at 13 04 16

### DIFF
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -67,7 +67,7 @@ a:focus {
   left: -9999em;
 
   /* Default link colour doesn't have enough contrast against $focus-colour */
-  &:focus, 
+  &:focus,
   &:visited {
     color: $text-colour;
   }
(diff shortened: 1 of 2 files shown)


robinwhittleton:
Why are we changing it here and not [where it’s originally defined](https://github.com/alphagov/govuk_template/blob/4eef85260b0ae1543fb7af8093ec30db0c98c3f3/source/assets/stylesheets/_header.scss#L205)?

edwardhorsford:
@robinwhittleton because it looked like other stuff related to focus was set in the accessibility file, including the existing line that was setting links to black.

At best it looks like we're setting anchors in the header to black twice?

robinwhittleton:
I’d rather not apply this where you’ve suggested. It should probably be an extra `:focus` state after the `#proposition-name:hover` block at https://github.com/alphagov/govuk_template/blob/4eef85260b0ae1543fb7af8093ec30db0c98c3f3/source/assets/stylesheets/_header.scss#L114 .

gemmaleigh:
I have tested these changes locally. 

mar-16-2017 11-43-00

PR #275, opened 2017-02-08
Allow skiplink target to be overridden

### LINKED ISSUE #271: Skip link is hardcoded to point at #content
GOV.UK Template hardcodes the skiplink to point at `#content` (see [this line](https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L39)). But GOV.UK Elements adds styles to `#content` (see [this block](https://github.com/alphagov/govuk_elements/blob/d6226bd2c8c7e256f995c0f519daaa4035c6bd54/public/sass/elements/_layout.scss#L9-L20)).

There are cases where we don’t want the thing we identify as `#content` to have those styles.

Would it be possible to override where the skiplink points to, so we could have `<a href="#anything-else" class="skiplink">`?

quis:
Fixes #271

---

GOV.UK Template hardcodes the skiplink to point at `#content`. But sometimes using `#content` introduces unwanted styling from 3rd parties (eg [in GOV.UK Elements](https://github.com/alphagov/govuk_elements/blob/d6226bd2c8c7e256f995c0f519daaa4035c6bd54/public/sass/elements/_layout.scss#L9-L20)).

This commit allows users of the template to set their own target for the skiplink. They will still need to include the target element in their pages, same as they have to for `id='content'` at the moment.

### DIFF
--- a/build_tools/compiler/django_processor.rb
+++ b/build_tools/compiler/django_processor.rb
@@ -31,6 +31,7 @@ module Compiler
       proposition_header:   block_for(:proposition_header),
       top_of_page:          "{% load staticfiles %}" + block_for(:top_of_page),
       skip_link_message:    statement_tag_for(:skip_link_message, 'Skip to main content'),
+      skip_link_target:     statement_tag_for(:skip_link_target, 'content'),
       logo_link_title:      statement_tag_for(:logo_link_title, 'Go to the GOV.UK homepage'),
       licence_message:      block_for(:licence_message, '<p>All content is available under the <a href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence v3.0</a>, except where otherwise stated</p>'),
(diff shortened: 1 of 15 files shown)


nickcolley:
Should we also split this into another issue to avoid styling based on skip links.

I'd propose:

- Change #content to #govuk-skip-link-target (or something very explicit)
- Change styling from #content to a specific presentational class.

quis:
@nickcolley That would cause breaking changes for consumers of Template and Elements. Feels overkill to fix my rather specific problem – even if the end result would be more robust/self-explanatory.

gemmaleigh:
@quis, would you be able to update this to fix the failing tests?

timpaul:
Following the launch of the [GOV.UK Design System](https://design-system.service.gov.uk/), GOV.UK Template will now only get major bug fixes and security patches, so I'm going to close this.

The issue that this PR addresses has been [captured on GOV.UK Frontend here](https://github.com/alphagov/govuk-frontend/issues/971).

Issue #277, opened 2017-02-20
Add support for theme-color?


willp-bl:
It would be nice to be able to have [theme-color](https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android) set so the browser window matches the site.  Robin and I tested this in the verify-frontend by adding `<meta name="theme-color" content="#0b0c0c" />`

That only works for Android.  iOS did not seem to have a similar option and we had not tested Windows Mobile.

Can this be added to govuk_template?

nickcolley:
Yeah this makes sense!

If you want to make a PR and need some help let us know. :)

PR #278, opened 2017-02-20
Add support for `theme-color`

willp-bl:
This is used in Chrome for Android, see:
https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android

See #277

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -17,12 +17,16 @@
     <!--[if lt IE 9]><script src="<%= asset_path "ie.js" %>"></script><![endif]-->
 
     <link rel="shortcut icon" href="<%= asset_path 'favicon.ico' %>" type="image/x-icon" />
+    <%# the colour used for mask-icon is the standard palette $black from 
+        https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss %>
     <link rel="mask-icon" href="<%= asset_path 'gov.uk_logotype_crown.svg' %>" color="#0b0c0c">
     <link rel="apple-touch-icon-precomposed" sizes="152x152" href="<%= asset_path "apple-touch-icon-152x152.png" %>">
     <link rel="apple-touch-icon-precomposed" sizes="120x120" href="<%= asset_path "apple-touch-icon-120x120.png" %>">
     <link rel="apple-touch-icon-precomposed" sizes="76x76" href="<%= asset_path "apple-touch-icon-76x76.png" %>">
     <link rel="apple-touch-icon-precomposed" href="<%= asset_path "apple-touch-icon-60x60.png" %>">
 
+    <%# the colour used for theme-color is the standard palette $black from 
+        https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss %>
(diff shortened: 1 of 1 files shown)


nickcolley:
Sweet thanks @willp-bl 👍 

Can you show some screenshots of it in action?

dsingleton:
Are we happy using a hard coded version of [a colour from our palette](https://github.com/alphagov/govuk_frontend_toolkit/blob/89168934d1455754f455c398ba5c350a484d9a2a/stylesheets/colours/_palette.scss#L50)? 

Looking at the template only it feels like a magic value. I think we'd want to include a comment at least, which shows where the value comes from, or ideally pull it in from the SCSS variable as part of the build process.

I don't want to slow down this PR too much, but we should consider how we'd keep this value in sync.

willp-bl:
I don't have access to BrowserStack, could someone lend a hand with the screenshots?

The hard coded colour is used a couple of lines above without a comment - https://github.com/alphagov/govuk_template/pull/278/files#diff-fe08901690b6fdc6f34614aaae900216R20

Happy to add a comment to explain to this PR but I don't know where it is from...

willp-bl:
Should have clicked that link, will add a comment

dsingleton:
@willp-bl Thanks! That feels much better. Could you use [`erb` style comments](https://github.com/willp-bl/govuk_template/blob/04d5071cd816cc1800eaef9086c7fd63b0cf3efa/source/views/layouts/govuk_template.html.erb#L113)? That will be useful if looking at the template source, but won't add to the page for end users.

source/views/layouts/govuk_template.html.erb
```diff
@@ -17,12 +17,18 @@
...
     <link rel="shortcut icon" href="<%= asset_path 'favicon.ico' %>" type="image/x-icon" />
+    <%# the colour used for mask-icon is the standard palette $black from 
+        https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss %>
     <link rel="mask-icon" href="<%= asset_path 'gov.uk_logotype_crown.svg' %>" color="#0b0c0c">
     <link rel="apple-touch-icon-precomposed" sizes="152x152" href="<%= asset_path "apple-touch-icon-152x152.png" %>">
     <link rel="apple-touch-icon-precomposed" sizes="120x120" href="<%= asset_path "apple-touch-icon-120x120.png" %>">
     <link rel="apple-touch-icon-precomposed" sizes="76x76" href="<%= asset_path "apple-touch-icon-76x76.png" %>">
     <link rel="apple-touch-icon-precomposed" href="<%= asset_path "apple-touch-icon-60x60.png" %>">
 
+    <%# the colour used for mask-icon is the standard palette $black from
```
fofr: `mask-icon` should be `theme-color` here

willp-bl: Thanks, just pushed a fix for that :)

dsingleton:
@willp-bl Thanks for changes! This is looking good to me.

I'm happy to merge unless anyone has objections?

nickcolley:
I'm just going to check this out on an android device but after that 👍

nickcolley:
Before:

image

After:

image

image

Looks even better without the cookie bar! Thanks Will 💯

PR #281, opened 2017-03-10
[Discuss] Add text-decoration-skip: ink to all links on GOV.UK

nickcolley:
Based on some design updates Mia Allers has made, will work in Chrome Canary soon in Chrome Stable and Safari.

[This prevents underlines from being drawn too close to glyphs](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip)

http://caniuse.com/#feat=text-decoration

I've included the `-webkit-` prefix but having trouble as it seems like some Safari versions already skip the underline, need to investigate more...

Edit: I've had a look and it seems Safari does this behaviour by default.

Be good to get some feedback on edge cases people can think where we would not want to do this globally.

| Before | After |
|--------|------|
|||
| | |
|||
| | |
|||

| Before | After |
|--------|------|
||screencapture-gov-uk-social-security-child-support-tribunal-1489161673155 |
| | |
||screencapture-gov-uk-1489161620081|
| | |
|||

### DIFF
--- a/source/assets/stylesheets/_basic.scss
+++ b/source/assets/stylesheets/_basic.scss
@@ -80,6 +80,7 @@ fieldset {
 
 a:link {
   color: $link-colour;
+  text-decoration-skip: ink;
 }
 
 a:visited {


robinwhittleton:
I’m certainly happy with this idea - it’s a single change that will make things look a little nicer in a few browsers and certainly no worse in the ones that don’t support it. For what it’s worth, the Mozilla bug to support it is https://bugzilla.mozilla.org/show_bug.cgi?id=812990; Microsoft don’t mention it [on their platform roadmap](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/).

36degrees:
Gets a 👍 from me too.

edwardhorsford:
Looks good 👍  - prefer these links.

Might be helpful to show a smaller screenshot? Took me a second to work out what the change was.

nickcolley:
Sorry @edwardhorsford you're right it's not so obvious. I've updated the original comment to be clearer.

nickcolley:
So looks like Safari does the `ink` behaviour by default so I'm thinking of maybe removing the `-webkit-` prefix line, thoughts?

36degrees:
I can't see anything on MDN or caniuse to suggest that `-webkit-text-decoration-skip: ink` does anything in any browser, so I think it can be removed.

fofr:
I think the typography in Chrome makes some of these links harder to read – looks like rogue full-stops.

## Chrome

screen shot 2017-05-05 at 11 37 38
screen shot 2017-05-05 at 11 37 32

## Safari

screen shot 2017-05-05 at 11 38 15
screen shot 2017-05-05 at 11 38 11

PR #283, opened 2017-03-17
Add block variable for body attributes

GreensterRox:
**What does it do?**

I've inheritied a project which relies on GA event tagging being driven by attributes dynamically added to the body tag.
Would you accept a modification which allows for a block variable being added to the body tag? This should be a non-breaking change for anyone not using this block var.

This change would allow uses of the template to specify custom attributes on the body tag if they so wish, the default would remain blank. We use it like this:

`<body class="{% block body_classes %}{% endblock %}" {% block body_attributes %}{% endblock %}>`

to render HTML like this:

`<body class="noscript"  data-journey="report-benefit-fraud:stage:details" >`

**What type of change is it?**

New feature (non-breaking change which adds functionality)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.19.3
+
+- Add block variable for body attributes
+
 # 0.19.2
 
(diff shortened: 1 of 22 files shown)


GreensterRox:
Found some issues - resubmitted new PR

PR #282, opened 2017-03-16
Version 0.20.0

robinwhittleton:
- Fix EJS template ([PR #270](https://github.com/alphagov/govuk_template/pull/270))
- Add `theme-color` support to make the page surround in Chrome’s tab view on Android match the black GOV.UK header ([PR #278](https://github.com/alphagov/govuk_template/pull/278))
- Add `text-decoration-skip: ink` to all links on GOV.UK ([PR #281](https://github.com/alphagov/govuk_template/pull/281))
- Improve contrast of links when focused ([PR #272](https://github.com/alphagov/govuk_template/pull/272))
- Make header text colour black when focused ([PR #274](https://github.com/alphagov/govuk_template/pull/274))

PR #285, opened 2017-03-23
Add block variable for body attributes

GreensterRox:
**What does it do?**

I've inheritied a project which relies on GA event tagging being driven by attributes dynamically added to the body tag.
Would you accept a modification which allows for a block variable being added to the body tag? This should be a non-breaking change for anyone not using this block var.

This change would allow uses of the template to specify custom attributes on the body tag if they so wish, the default would remain blank. We use it like this:

`<body class="{% block body_classes %}{% endblock %}" {% block body_attributes %}{% endblock %}>`

to render HTML like this:

`<body class="noscript" data-journey="report-benefit-fraud:stage:details" >`

**What type of change is it?**

New feature (non-breaking change which adds functionality)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ You can [view a collection of auto-generated examples](http://alphagov.github.io
 
 ## Requirements
 
-The Ruby language (1.9.3+), the build tool [Rake](http://rake.rubyforge.org/) & the dependancy management tool [Bundler](http://bundler.io/)
+The Ruby language (1.9.3+), the build tool [Rake](http://rake.rubyforge.org/), the dependency management tool [Bundler](http://bundler.io/) & [Java](https://www.java.com/en/) 
 
 ## Detailed Docs
 
--- a/build_tools/compiler/django_processor.rb
+++ b/build_tools/compiler/django_processor.rb
@@ -15,6 +15,7 @@ module Compiler
     @@yield_hash = {
       after_header:         block_for(:after_header),
       body_classes:         block_for(:body_classes),
+      body_attributes:      block_for(:body_attributes),
       body_start:           block_for(:body_start),
       body_end:             block_for(:body_end),
       content:              block_for(:content),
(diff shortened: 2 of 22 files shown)


36degrees:
Thanks for your PR!

Versioning is done separately to individual 'feature' pull requests – like [this](https://github.com/alphagov/govuk_template/pull/282) – so please revert your changes to `CHANGELOG.md` and `lib/govuk_template/version.rb`.

GreensterRox:
Hi @36degrees , thanks for your message. When I revert these two files do you still need me to squash into a single commit id ?

36degrees:
That'd be preferred, please. It just helps keep the repo's history clean.

GreensterRox:
I had to create another PR [https://github.com/alphagov/govuk_template/pull/286] because I can't seem to squash commits once a branch is pushed to the server - unless you know a way of doing this ?

36degrees:
What's the error message you're seeing? You'll need to force push as you're rewriting history.

GreensterRox:
I did force push and that seems to have done the trick :-)

README.md
```diff
@@ -13,7 +13,7 @@ You can [view a collection of auto-generated examples](http://alphagov.github.io
 
 ## Requirements
 
-The Ruby language (1.9.3+), the build tool [Rake](http://rake.rubyforge.org/) & the dependancy management tool [Bundler](http://bundler.io/)
```
robinwhittleton: You’re not adding anything that requires Java though? Or is it already a requirement that’s been missed?

GreensterRox: Hi @robinwhittleton, I noticed that 'webjar_packager.rb' attempts to create a jar file. That's why I added Java as a requirement. I had to install Java to make this work.

fofr:
This PR seems to have stalled.

@GreensterRox do you still need this feature?

@36degrees Is there anything specific holding up this PR?

Is this a feature we want to add?

GreensterRox:
Hi @fofr, yes, this is indeed a feature that would help the project that I was previously working on, as it gives some extra flexibility by allowing custom attributes to be defined on the body tag.

timpaul:
Thank you for this work.

We have captured [the need for this feature in GOV.UK Frontend here](https://github.com/alphagov/govuk-frontend/issues/971).

PR #290, opened 2017-03-27
Link to https national archives

edwardhorsford:
Change the national archives link to https rather than http

With thanks to @8032 for reporting.

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -96,7 +96,7 @@
           </div>
 
           <div class="copyright">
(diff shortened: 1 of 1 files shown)


nickcolley:
Seems like a good change, the national archives themselves have mixed content warnings so wont actually be guaranteed to be secure for users going there.

Not sure if we know anyone over there to poke.

screen shot 2017-03-27 at 14 19 41

edwardhorsford:
Based on some sleuthing by @davidillsley it looks like this [may be intentional](https://github.com/alphagov/govuk_template/commit/091f855551923733c9e7df11f0eb66660653fa79).

davidillsley:
The national archives seem to have (almost completely) fixed their mixed content problem - a single insecure png remains, so I don't think there's any argument to not do this any more

Issue #238, opened 2016-08-16
Integrate SRI for scripts / stylesheets


robinwhittleton:
The [Subresource Integrity specification](https://www.w3.org/TR/SRI/) adds a new `integrity` attribute to `<script>` and `<link>` elements that is a hash of the asset they point to. Browsers that support SRI can then hash the downloaded asset and check that it matches the hash specified in the integrity attribute. We [recently implemented this](https://github.com/alphagov/verify-frontend/pull/192) on [verify-frontend](https://github.com/alphagov/verify-frontend/) for the application specific files but we’d like to turn it on for the govuk_template styles and scripts as well.

On verify-frontend it was as simple as using [sprockets-rails](https://github.com/rails/sprockets-rails) `integrity` attribute passed into `javascript_include_tag` and `stylesheet_link_tag` but as we compile govuk_template into multiple different template languages we can’t do that here. To implement this we’ll have to hash [govuk-template.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/govuk-template.js) and [govuk-template.scss](https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/govuk-template.scss) post-compilation, and add those hashes manually to an `integrity` attribute on their respective `<script>` and `<link>` tags.

robinwhittleton:
Got a branch that’ll generate integrity hashes for tagged `<link>`s and ’<script>`s and compile them into the template at https://github.com/alphagov/govuk_template/tree/sri . Still to do: fix the tests and add some more around this functionality, and check to make sure that everything is still being packaged correctly.

boffbowsh:
We have a ticket in Core Formats' backlog to look at this https://trello.com/c/58oo6NSQ/481-spike-enable-subresource-integrity-sri-on-our-frontend-apps-1-day

robinwhittleton:
Yeah, that’s for the individual frontend services (or rather, one to start). Hopefully on those it should be trivial, or [at least it was for Verify](https://github.com/alphagov/verify-frontend/pull/192). That only covers the JS/CSS supplied in the individual apps though. To gain full coverage GOV.UK will need to enable SRI in static (again, fairly trivial once you’ve verified that the approach is OK) and to have it in the template which is what this work does. Doing this also means that verify-frontend will have SRI applied to every CS/JSS asset - currently it’s not on the template assets.

robinwhittleton:
[...]

PR #294, opened 2017-04-25
[DO NOT MERGE] SRI

elenatanasoiu:
[Trello card](https://trello.com/c/oFhtO0Gm/146-enable-subresource-integrity-sri-on-government-frontend-l)

[Github Issue](https://github.com/alphagov/govuk_template/issues/238)

This PR will generate integrity hashes for tagged <link>s and <script>s and compile them into the template. 

The [Subresource Integrity specification](https://www.w3.org/TR/SRI/) adds a new `integrity` attribute to `<script>` and `<link>` elements that is a hash of the asset they point to. Browsers that support SRI can then hash the downloaded asset and check that it matches the hash specified in the integrity attribute. This [was recently implemented](https://github.com/alphagov/verify-frontend/pull/192) on [verify-frontend](https://github.com/alphagov/verify-frontend/) for the application specific files, and is now being turned on for the govuk_template styles and scripts as well.

On verify-frontend it was as simple as using [sprockets-rails](https://github.com/rails/sprockets-rails) `integrity` attribute passed into `javascript_include_tag` and `stylesheet_link_tag` but as we compile govuk_template into multiple different template languages we can’t do that here. To implement this we’ll have to hash [govuk-template.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/govuk-template.js) and [govuk-template.scss](https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/govuk-template.scss) post-compilation, and add those hashes manually to an `integrity` attribute on their respective `<script>` and `<link>` tags.

Still to do: fix the tests and add some more around this functionality, and check to make sure that everything is still being packaged correctly.

Depends on: 
- government-frontend: https://github.com/alphagov/government-frontend/pull/331
- static: https://github.com/alphagov/static/pull/1008

### DIFF
--- a/build_tools/compiler/asset_compiler.rb
+++ b/build_tools/compiler/asset_compiler.rb
@@ -102,7 +102,7 @@ module Compiler
           modified = File.open(original.path + '.tmp', 'w')
           contents = original.read
           @integrity_attributes.each do |key, value|
-            asset_expression = Regexp.new "#{key}(.*?)integrity=\"\"(.*?)>"
+            asset_expression = Regexp.new "#{key}(.*?integrity=\")(\".*?)>"
             contents.gsub! asset_expression, "#{key}\\1#{value}\\2>"
           end
           modified.puts contents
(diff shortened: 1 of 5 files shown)


elenatanasoiu:
This PR is not going to be used anymore, so I'm closing it. Arguments for closing this PR can be found here: https://github.com/alphagov/govuk_template/pull/301

PR #297, opened 2017-05-09
Add publisher to build_tools for gems

tombye:
### The problem

The Gem Publisher library runs git commands to tag
the version its publishing. It assumes creds for
this are already set up but, in the switch from
Jenkins to Travis, this is no longer the case.

### How does this solve it?

This creates a local publisher, following the
interface of the ones we already have, and
wrapping our use of Gem Publisher.

This means we can set up the github creds in the
same way as the other publishers, using an
existing token.

### Why call it ErbPublisher?

The new publisher calls Gem Publisher so there's a
name conflict if it is also called GemPublisher.

Caveat: I'm no rubyist so there may be better
approaches to this but I figured following the
convention the other publishers had, of being
named after their templating language would at
least be sensible.

### DIFF
--- a/build_tools/packager/gem_packager.rb
+++ b/build_tools/packager/gem_packager.rb
@@ -18,7 +18,7 @@ module Packager
       end
     end
 
-    private
+  private
 
     def run(command)
       output, status = Open3.capture2e(command)
--- a/build_tools/packager/gem_packager.rb
+++ b/build_tools/packager/gem_packager.rb
@@ -1,5 +1,9 @@
+require_relative '../helpers'
+
 module Packager
   class GemPackager
+    include Helpers
+
     def self.build
       new.build
     end
@@ -17,13 +21,5 @@ module Packager
         FileUtils.mv(gem, "pkg")
       end
     end
-
-  private
-
-    def run(command)
-      output, status = Open3.capture2e(command)
-      abort "Error running #{command}: exit #{status.exitstatus}\n#{output}" if status.exitstatus > 0
-      output
-    end    
   end
 end
--- a/build_tools/packager/gem_packager.rb
+++ b/build_tools/packager/gem_packager.rb
@@ -1,5 +1,5 @@
 module Packager
-  class ErbPackager
+  class GemPackager
     def self.build
       new.build
     end
--- a/build_tools/publisher/gem_publisher.rb
+++ b/build_tools/publisher/gem_publisher.rb
@@ -1,10 +1,11 @@
 require 'govuk_template/version'
+require_relative '../helpers'
 require 'tmpdir'
 require 'open3'
 
 module Publisher
   class GemPublisher
-    attr_accessor :version
+    include Helpers
 
     GIT_URL = "https://#{ENV['GITHUB_TOKEN']}@github.com/alphagov/govuk_template.git"
 
@@ -22,19 +23,6 @@ module Publisher
           run "git push --tags origin master"
         end
       end
-    end
-
-    def version_released?
-      output = run "git ls-remote --tags #{GIT_URL.shellescape}"
-      return !! output.match(/v#{@version}/)
-    end
-
-  private
-
-    def run(command)
-      output, status = Open3.capture2e(command)
-      abort "Error running #{command}: exit #{status.exitstatus}\n#{output}" if status.exitstatus > 0
-      output
     end
   end
 end
--- a/build_tools/publisher/gem_publisher.rb
+++ b/build_tools/publisher/gem_publisher.rb
@@ -29,7 +29,7 @@ module Publisher
       return !! output.match(/v#{@version}/)
     end
 
-    private
+  private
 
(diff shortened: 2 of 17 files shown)


36degrees:
Is gem_publisher [this](https://github.com/gds-attic/gem_publisher)? It seems odd to continue to rely on / extend a thing which has been archived to the attic…

tombye:
That's a good point, I'll see if there are any alternatives. Do you know of any?

36degrees:
Have we explored making use of Travis' [built in gem releasing](https://docs.travis-ci.com/user/deployment/rubygems/)?

tombye:
No, excellent tip, will investigate!

tombye:
I've done some investigation and I think there's two potential pieces of work from this discussion:

### Removing GemPublisher from this repo

GemPublisher is used to:
1. [build the gem](https://github.com/alphagov/govuk_template/blob/master/build_tools/packager/gem_packager.rb#L13)
2. [determine if there is a tag for a gem version](https://github.com/alphagov/govuk_template/blob/master/Rakefile#L99)
3. [publish and tag the release](https://github.com/alphagov/govuk_template/blob/master/Rakefile#L103)

### Using travis to publish the gem

This would require:
1. encrypting the API key and adding to `.travis.yml`
2. removing the existing key file from `.travis/secrets.tar.enc`

I don't think it'd be too much work to remove GemPublisher but as a newbie to ruby I'd be interested in where to put the code that would still be needed to do the jobs listed in 'Removing GemPublisher from this repo' above. They don't seem to fit into the current structure in `build_tools` and I'm not sure what is a good convention.

fofr:
@tombye Should this PR be closed? Sounds like it needs some follow-up work, one way or another?

tombye:
@36degrees I've removed all use of gem_publisher from the packager, publisher and Rakefile.

I also looked into moving the publishing into the `.travis.yml` file. 

This would work for publishing the gem as this repo *is* the gem but the NPM packages need to be built (into the `pkg` folder) before they are published). In other words the rake task for publishing the release would need re-working.

I plan to do that work but keep it separate from this pull request so releases aren't blocked and we have a working build to refactor from.

@fofr I was a bit delayed doing this because my time is split between projects so I could only work on this a few days a week.

build_tools/publisher/gem_publisher.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def initialize(version = GovukTemplate::VERSION)
+      @version = version
+    end
+
+    def publish
+      puts "Pushing govuk_template-#{GovukTemplate::VERSION}"
+      run "gem push pkg/govuk_template-#{GovukTemplate::VERSION}.gem"
+      Dir.mktmpdir("govuk_template_gem") do |dir|
+        run "git clone -q #{GIT_URL.shellescape} #{dir.shellescape}"
```
36degrees: Will this output the GitHub token (as part of `GIT_URL`) if the clone fails?

build_tools/packager/gem_packager.rb
```diff
@@ -10,11 +10,20 @@ def initialize
...
-        gem = GemPublisher::Builder.new.build('govuk_template.gemspec')
+        stdout_str = run('gem build govuk_template.gemspec')
+        gem = stdout_str[/File:\s+(.+)/, 1]
 
         @repo_root.join('pkg').mkpath
         FileUtils.mv(gem, "pkg")
       end
     end
+
+    private
```
36degrees: GDS style is [to outdent `private`](https://github.com/alphagov/service-manual-publisher/blob/master/app/forms/guide_form.rb#L8)

build_tools/publisher/gem_publisher.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def version_released?
+      output = run "git ls-remote --tags #{GIT_URL.shellescape}"
+      return !! output.match(/v#{@version}/)
+    end
+
+    private
+
+    def run(command)
```
36degrees: Worth abstracting this as it's duplicated across the two classes?

tombye: There are a few of these in `build_tools` so I think that would be good.

build_tools/packager/gem_packager.rb
tombye: 👍

build_tools/publisher/gem_publisher.rb
tombye: Excellent catch 👏

I've run a local test of this, simulating a bad connection to the repository domain and it sends this to `stderr`:

```
fatal: unable to access 'https://*ENV*@github.com/alphagov/govuk_template.git/': Failed to connect to github.com port 443: Connection refused
```
(where \*ENV\* is the github token.)

That being the case, I've rewritten how we run shell commands in 146aa55, to allow custom errors and used them to mask this kind of output.

PR #301, opened 2017-05-12
Add SRI

elenatanasoiu:
[Trello card](https://trello.com/c/oFhtO0Gm/146-enable-subresource-integrity-sri-on-government-frontend-l)

[Github issue](https://github.com/alphagov/govuk_template/issues/238)

Based on a previous attempt to add SRI - see PR [here](https://github.com/alphagov/govuk_template/pull/294)

## What this does

This PR enables govuk_template to use SRI for the css and js assets it is serving to other projects.

### What is [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)

SRI will add an `integrity` attribute on your script tags:

```
<script src="https://example.com/example.css" 
integrity="sha384oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8w" 
crossorigin="anonymous"></script>
```
The example above is generated automatically by sprockets-rails in your project if you do this: 
`<%= stylesheet_script_tag 'example', integrity: true %>`

The way the digest value for the `integrity` attribute is calculated is by applying SHA256 on the **contents** of the file. In that way, the browser can double check that the right file is being received and hasn’t been tampered with (which would mean the contents would have changed and the resulting digest would be incorrect).

### The problem with using SRI in govuk_template
The css files provided by govuk_template (for example: to the static project) are still `css.erb` files. In which case the contents of those files will change when they are processed in static (specifically there’s multiple `<%= asset_path(...)%>`s in there).

So here’s the problem illustrated:

1. The `govuk_template` gem compiles a file called `govuk_template.css.erb` which contains multiple `asset_path` lines, e.g.
```background-image: url(<%= asset_path 'images/govuk-crest.png' %>);```

2. static receives this file and replaces all the asset_path lines with it’s own path:
```background-image: url(**http://static.dev.gov.uk/static**/images/govuk-crest-2x.png);```
Notice that this url will vary depending on the app that uses the gem. As a result, the contents of the resulting `govuk_template.css` will vary depending on which app is using the gem so the digest will also vary.
This means you can’t generate the digest inside the gem, and the calculation will have to be left to the apps to do individually.

3. You can't use sprockets-rails in govuk_template directly because the assets it serves are also compiled into django, play, liquid, mustache and other crazy formats.
So you can’t do this:
`<%= stylesheet_link_tag ‘govuk-template-print’, integrity: true%>`
and then leave it up to sprockets-rails to calculate the digest for you, because this will not work for django and all the other formats. They don’t know what `stylesheet_tag` is.  
Instead, this is what govuk_template does so it can be compiled into multiple languages:
` <link href="<%= asset_path "govuk-template-print.css" %>" media="print" rel="stylesheet" />`

4. You can't calculate the integrity digest by hand in the gem, because of point 2 -> the content of the file will change depending on which application is using it. 

## Proposed solution

We use `stylesheet_link_tag` and set `integrity: true`, but we leave it up to the apps using the gem to actually generate the integrity digest. For the other languages (django, jijna, liquid, mustache) we define a `stylesheet_include_tag` method in their respective processors that will translate that tag into something they can understand:
`<link href="<%= asset_path "govuk-template-print.css" %>" media="print" rel="stylesheet" />`

### DIFF
--- a/build_tools/compiler/template_processor.rb
+++ b/build_tools/compiler/template_processor.rb
@@ -1,4 +1,6 @@
 require 'erb'
+require 'active_support/core_ext/hash'
+require 'active_support/core_ext/array'
 
 module Compiler
   class TemplateProcessor
@@ -41,7 +43,7 @@ module Compiler
     end
 
     def stylesheet_link_tag(*sources)
-      options = stringify_keys(extract_options!(sources))
+      options = exclude_sri_fields(sources.extract_options!)
       sources.uniq.map { |source|
         link_options = {
             "rel" => "stylesheet",
@@ -53,7 +55,7 @@ module Compiler
     end
 
     def javascript_include_tag(*sources)
-      options = stringify_keys(extract_options!(sources))
+      options = exclude_sri_fields(sources.extract_options!)
       sources.uniq.map { |source|
         script_options = {
             "src" => asset_path(source)
@@ -62,32 +64,16 @@ module Compiler
       }.join("\n")
     end
 
+    def exclude_sri_fields(options)
+      options.stringify_keys.except("integrity", "crossorigin")
+    end
+
     def content_tag(name, options = nil)
       "<#{name}#{options}></#{name}>"
     end
 
     def tag(name, options)
       "<#{name}#{options}/>"
-    end
-
-    def extract_options!(options)
-      if options.last.is_a?(Hash) && extractable_options?(options.last)
-        options.pop
-      else
-        {}
-      end
-    end
-
-    def extractable_options?(options)
-      options.instance_of?(Hash)
-    end
-
-    def stringify_keys(options)
-      result = {}
-      options.each_key do |key|
-        result[key.to_s] = options[key]
-      end
-      result
     end
 
     def tag_options(options)
--- a/build_tools/compiler/template_processor.rb
+++ b/build_tools/compiler/template_processor.rb
@@ -39,5 +39,69 @@ module Compiler
     def method_missing(name, *args)
       puts "#{name} #{args.inspect}"
     end
+
+    def stylesheet_link_tag(*sources)
+      options = stringify_keys(extract_options!(sources))
+      sources.uniq.map { |source|
+        link_options = {
+            "rel" => "stylesheet",
+            "media" => "screen",
+            "href" => asset_path(source)
+        }.merge!(options)
+        tag(:link, tag_options(link_options))
+      }.join("\n")
+    end
+
+    def javascript_include_tag(*sources)
+      options = stringify_keys(extract_options!(sources))
+      sources.uniq.map { |source|
+        script_options = {
+            "src" => asset_path(source)
+        }.merge!(options)
+        content_tag(:script, tag_options(script_options))
+      }.join("\n")
+    end
+
+    def content_tag(name, options = nil)
+      "<#{name}#{options}></#{name}>"
+    end
+
+    def tag(name, options)
+      "<#{name}#{options}/>"
+    end
+
+    def extract_options!(options)
+      if options.last.is_a?(Hash) && extractable_options?(options.last)
+        options.pop
+      else
+        {}
+      end
+    end
+
+    def extractable_options?(options)
+      options.instance_of?(Hash)
+    end
+
+    def stringify_keys(options)
+      result = {}
+      options.each_key do |key|
+        result[key.to_s] = options[key]
+      end
+      result
+    end
+
+    def tag_options(options)
+      return if options.empty?
+      output = "".dup
+      sep    = " "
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: build_tools/compiler/template_processor.rb [hunk 1]
```diff
@@ -84,8 +85,10 @@ def extractable_options?(options)
 
     def stringify_keys(options)
       result = {}
-      options.each_key do |key|
-        result[key.to_s] = options[key]
+      if !options.empty?
```
h-lame: minor nitpick: do we need this conditional?  if options is `.empty?` then the `.each_key` iterator won't do anything, so it should be safe to leave this method as it was.

build_tools/compiler/template_processor.rb
```diff
@@ -39,5 +40,71 @@ def asset_path(file, options={})
...
+      else
+        {}
+      end
+    end
+
+    def extractable_options?(options)
+      options.instance_of?(Hash)
+    end
+
+    def stringify_keys(options)
```
h-lame: Because we rely on 'active_support/core_ext/hash' we can call `.stringify_keys` on the hash directly, so we don't need to implement this here.  It's in rails 3.1 so doesn't break our dependency (see: http://guides.rubyonrails.org/v3.1/active_support_core_extensions.html#stringify_keys-and-stringify_keys).

build_tools/compiler/template_processor.rb
```diff
@@ -39,5 +40,71 @@ def asset_path(file, options={})
...
+
+    def content_tag(name, options = nil)
+      "<#{name}#{options}></#{name}>"
+    end
+
+    def tag(name, options)
+      "<#{name}#{options}/>"
+    end
+
+    def extract_options!(options)
```
h-lame: If we added a require for `activesupport/core/ext/array` as well as the hash one then we could also rely on `Array#extract_options!` and not need to implement it here.  It's also in rails 3.1 and wouldn't break our dependency (see: http://guides.rubyonrails.org/v3.1/active_support_core_extensions.html#options-extraction)

elenatanasoiu:
sprockets-rails was removed from rails >= 4, so it won't make sense to add a dependency to sprockets-rails if your version of rails is still 3.1. 

I've added details about how SRI can be used with `govuk_template` gem in the docs

spec/support/examples/processor.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      it "should parse the hash" do
+        expect(processor.tag_options(options)).to eql(" rel=\"stylesheet\" media=\"screen\" href=\"#{processor.asset_path(css_source)}\"")
+      end
+    end
+
+    describe "#exclude_sri_fields" do
+      let(:options) { {"rel"=>"stylesheet", "media"=>"screen", "href"=>processor.asset_path(css_source), "integrity" => true, "crossorigin" => "anonymous" } }
+
+      it "should parse the hash" do
```
h-lame: "should parse the hash" could be more descriptive.  "removes integrity and crossorigin keys from the hash" perhaps?

spec/support/examples/processor.rb
```diff
@@ -0,0 +1,80 @@
...
+    describe "#tag" do
+      it "should return the correct html link tag" do
+        expect(processor.tag(:link, " rel=\"stylesheet\" media=\"screen\" href=\"#{processor.asset_path(css_source)}\"")).to eql("<link rel=\"stylesheet\" media=\"screen\" href=\"#{processor.asset_path(css_source)}\"/>")
+      end
+    end
+
+    describe "#tag_options" do
+      let(:options) { {"rel"=>"stylesheet", "media"=>"screen", "href"=>processor.asset_path(css_source) } }
+
+      it "should parse the hash" do
```
h-lame: "should parse the hash" could be more descriptive.  "flattens the hash into a string of quoted html attributes" perhaps?

elenatanasoiu: Renamed

spec/support/examples/processor.rb
elenatanasoiu: Renamed

fofr:
Do we need to back this out?

h-lame:
@fofr maybe we should back this out and re-release, but I'd say we don't need to yank old versions as it's not broken or a security risk or anything like that.  Before we do anything though we need to investigate the impact more; it might be we're comfortable with the % of users it would affect and want to remerge the SRI stuff on government frontend.

fofr:
Bug affects users of Firefox versions less than 52, which is about 0.7% of all users in the last month, about 315k users. There's also a proportion of users stuck on version 45 because they can't upgrade.

screen shot 2017-05-22 at 10 14 50

PR #305, opened 2017-05-16
Adds new icons to precompile list

elenatanasoiu:
Following this commit: https://github.com/alphagov/govuk_template/commit/82d9459596581623ca70f8d415a9117026c20414

The icons will also have to be precompiled in order to be served to other projects.


h-lame:
Removed DNM as this is good to merge.

PR #307, opened 2017-05-19
Bump to 0.21.1

h-lame:
To release fix in #305


fofr:
Maybe relevant: https://github.com/alphagov/govuk_template/pull/301#issuecomment-303017212

h-lame:
Closing because we need to do a bigger version bump to account for #308.

PR #308, opened 2017-05-22
Revert #301 Add SRI

h-lame:
This is a manual revert of #301, as the PR could not be reverted automatically.
We're reverting this PR because there is a bug in the SRI implementation
of Firefox versions upto 52 which at time of writing accounts for 0.7% of
total traffic (~315k users).  We still want to implement SRI, but for now
we're holding off until we'd impact fewer users.


source/views/layouts/govuk_template.html.erb
```diff
@@ -6,16 +6,15 @@
...
+    <!--[if gt IE 8]><!--><link href="<%= asset_path "govuk-template.css" %>" media="screen" rel="stylesheet" /><!--<![endif]-->
     <!--[if IE 6]><link href="<%= asset_path "govuk-template-ie6.css" %>" media="screen" rel="stylesheet" /><![endif]-->
     <!--[if IE 7]><link href="<%= asset_path "govuk-template-ie7.css" %>" media="screen" rel="stylesheet" /><![endif]-->
     <!--[if IE 8]><link href="<%= asset_path "govuk-template-ie8.css" %>" media="screen" rel="stylesheet" /><![endif]-->
-    <%= stylesheet_link_tag "govuk-template-print.css", media: "print", integrity: true, crossorigin: "anonymous" %>
+    <link href="<%= asset_path "govuk-template-print.css" %>" media="print" rel="stylesheet" />
 
     <!--[if IE 8]><link href="<%= asset_path "fonts-ie8.css" %>" media="all" rel="stylesheet" /><![endif]-->
-    <!--[if gte IE 9]><!--><%= stylesheet_link_tag "fonts.css", media: "all", integrity: true, crossorigin: "anonymous" %><!--<![endif]-->
-    <!--[if gte IE 9]><!--><%= stylesheet_link_tag "fonts.css", media: "all", integrity: true, crossorigin: "anonymous" %><!--<![endif]-->
```
fofr: Looks like this was duplicated accidentally in original?

source/views/layouts/govuk_template.html.erb
```diff
@@ -6,16 +6,15 @@
...
-    <!--[if lt IE 9]><%= javascript_include_tag "ie.js", integrity: true, crossorigin: "anonymous" %><![endif]-->
+    <!--[if gte IE 9]><!--><link href="<%= asset_path "fonts.css" %>" media="all" rel="stylesheet" /><!--<![endif]-->
+    <!--[if lt IE 9]><script src="<% asset_path "ie.js" %>"></script><![endif]-->
```
fofr: asset_path is missing an equals here, should be: `<%=`

fofr: Spotted via this comparison: https://github.com/alphagov/govuk_template/compare/10d23486f66367131b6ba31521b28099820bbf21...revert-add-sri

h-lame: oops.  I blame having to make this change 4 times because I kept editing the built version instead of the source!  Thanks for spotting.

source/views/layouts/govuk_template.html.erb
h-lame: Yeah, I think so.

source/views/layouts/govuk_template.html.erb
h-lame: Fixed in a rebase.

PR #309, opened 2017-05-22
Bump to 0.22.0

h-lame:
Includes #308 and #305

### DIFF
--- a/lib/govuk_template/version.rb
+++ b/lib/govuk_template/version.rb
@@ -1,3 +1,3 @@
 module GovukTemplate
-  VERSION = "0.21.0"
+  VERSION = "0.22.0"
 end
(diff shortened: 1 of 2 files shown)


lib/govuk_template/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukTemplate
-  VERSION = "0.21.0"
+  VERSION = "0.21.1"
```
fofr: This is not 0.22

h-lame: I was so focussed on making sure I didn't miss the CHANGELOG entry from the other PR I didn't do the obvious thing!  Thanks for checking.  I've fixed it in a rebase.

PR #310, opened 2017-05-24
[WIP] [Do not merge!] Make 'global-cookie-message' optional 

paulmsmith:
**Why do this?**

So I'm working on a service that started life as a citizen facing service but eventually became an internal tool to be used by staff in job centres.

The cookie message is not required for internal tools (I believe/hope) which is good because we have an issue that the 'thin clients' used by DWP staff, clear cookies as users log in and out of their machines (lots of which are hot-desks).

I've also found that ReadWrite Gold 11 in a certain mode is reading aloud a dismissed cookie message on each page (could be poor implementation, need to investigate)

So we'd like to remove it, users don't need to see it and we've observed some dismissing it each time they sign into their machine and use the tool. (repetitive annoyance).

I'm opening this pull request because I'm happy to do the work but it might be that it's larger than 'just making the cookie message optional'.

**What needs to happen?**

- [ ] Update the compilers so that they can convert 'if' or 'unless' statements in ERB to the other templating languages. Is this possible? cc: @gemmaleigh 
- [ ] Update the `source/views/layouts/govuk_template.html.erb` to make the cookie message div optional. 
- [ ] Update the tests
- [ ] Build/Package the template(s)

Hopefully this starts the ball rolling.


tombye:
I've had a look at this. I'm also no rubyist but I think we could do something to make the optional sections of the template carry across to all the variants.

While the template doesn't currently support sections being conditional outside of the .erb version, I think we could do this in the same way as it handles optional text. In this case, if text is not provided it will use a default value: 

https://github.com/alphagov/govuk_template/blob/929bb6f759e45f4f130f966ee1980d6b01fcc885/build_tools/compiler/ejs_processor.rb#L21

If we do this for the cookie message, we should also do if for the `@omit_header` section.

Happy to help with this btw.

garyboyle:
If we are doing this would we exclude the [cookie-bar.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/cookie-bar.js) file for completeness?

And do without lines 4-8 of [core.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/core.js)?

And maybe even exclude the [cookie-functions.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/cookie-functions.js) file if we don't need cookie functionality?

Unfortunately I have no idea how this would be done.

paulmsmith:
Good question @garyboyle. I'd say that would be ideal but quite a complicated change to make to 'govuk_template'. 

[This Sprockets manifest](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/govuk-template.js) shows those cookie assets being included the in the build. 

So to remove them, we'd to make it possible for teams can opt-in/opt-out of such dependancies when consuming 'gov_template'. 

Short term, because those modules are coded defensively by [checking for the existing of the cookie message](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/cookie-bar.js#L7) then it's no huge issue for those to be included as it would have negligible impact on performance/size of the JS. 

Long term, making changes to this repository so that local configuration/build of 'govuk_template' is easier might be a good thing to do.

36degrees:
Thanks for your PR, @paulmsmith.

PR #314, opened 2017-06-09
Use newer versions of Node and NPM

36degrees:
Travis CI installs Node v0.10.36 with NPM v1.4.28 by default, which is really old and doesn't support using authentication tokens from ~/.npmrc, so we need to install something newer for npm publishing to work. Using Node 6.x as it is the latest LTS release.


36degrees:
Ideally need to remove the 0.22.1 tag from govuk_template_mustache repo before merging this, so that we go back and publish the mustache module.

gemmaleigh:
@36degrees this looks good to me, have you deleted the tag you mentioned? If so I'll merge.

gemmaleigh:
https://github.com/alphagov/govuk_template_mustache/releases is showing 0.22.1.

36degrees:
No, I don't have write access to that repo.

36degrees:
Tag deleted, merging now!

Issue #313, opened 2017-06-08
Consider removing link ink skipping


edwardhorsford:
In #281 we added `text-decoration-skip: ink;` to skip underlines near descenders - in theory because it looks better.

I wonder if we should reconsider this decision.

There are many examples where it makes single links look like multiple links. Are we happy with this? 

NB - it generally looks better in Safari which doesn't skip as much.

Discuss.

## Chrome

![screen shot 2017-05-05 at 11 37 38](https://cloud.githubusercontent.com/assets/319055/25742411/45c34f4c-3187-11e7-965c-86bf82ee1a72.png)
![screen shot 2017-05-05 at 11 37 32](https://cloud.githubusercontent.com/assets/319055/25742412/45c7f880-3187-11e7-9781-db70ac70cfe7.png)
![screen shot 2017-05-05 at 11 40 08](https://cloud.githubusercontent.com/assets/319055/25742496/a61a8a5e-3187-11e7-8272-01426d469ba0.png)

## Safari

![screen shot 2017-05-05 at 11 38 15](https://cloud.githubusercontent.com/assets/319055/25742434/58eade64-3187-11e7-9eba-13c6bb722730.png)
![screen shot 2017-05-05 at 11 38 11](https://cloud.githubusercontent.com/assets/319055/25742436/58ee3f32-3187-11e7-932a-0b484e4c5c76.png)

selfthinker:
I would be in favour of reverting it again. Evidence from a single user: It actually has happened to me that I opened the same page twice because I didn't realise that it was the same link.

soniaturcotte:
Agree. 

Additionally, there are also instances within single words where the letter combinations just look strange. 
<img width="422" alt="screen shot 2017-06-08 at 16 19 09" src="https://user-images.githubusercontent.com/19776089/26938960-b6296e0a-4c6d-11e7-8ecd-fb57db53e0fa.png">

alextea:
To the contrary in 4 rounds of research on GOV.UK since this was deployed we haven't seen any issues with people getting confused.

edwardhorsford:
I wonder if we might be able to test anything in analytics related to this - any ideas?

Perhaps: number of people opening the _same_ link twice from one source page - ie clicking both parts because they thought they were two links.

alextea:
Will discuss with Vin when he is back from leave as we are already measuring this stuff for navigation. Not sure if we will have any significant findings however.

selfthinker:
[...]

PR #316, opened 2017-06-14
Revert Add text-decoration-skip: ink to all links on GOV.UK

nickcolley:
We added underline skipping to hopefully make some content easier to read but looking
at some examples as seen at https://github.com/alphagov/govuk_template/issues/313 Chrome does a poor job of rendering underlines resulting in potentially confusing underlines that look like punctuation

This revert puts our links back to the 'boring' state of doing nothing special.


gemmaleigh:
Fixes #313.

PR #318, opened 2017-07-10
Fix negative offset on crown SVG causing left blob to be chopped

andymantell:


nickcolley:
Nice! Can you show some screenshots? 👍

andymantell:
Before:

image

After:

image

andymantell:
@nickcolley It looks like it's actually only used in Safari's pinned tabs though which is rendered so so small that it's imperceptible! The only reason I noticed is I was looking for an SVG copy of the crown to go in a PDF...

nickcolley:
Looks sounds as a 💷  👍

andymantell:
:+1:

PR #319, opened 2017-07-12
Fix django package

robyoung:
This is made up of two parts.

## Include assets in Django Python package

setuptools will not look for a `MANIFEST.in` unless the `include_package_data` flag is set to `True`.

This is described in the [setuptools documentation](http://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=include_package_data#including-data-files)

And in a slightly easier to understand way in the [Flask docs](http://flask.pocoo.org/docs/0.12/patterns/distribute/#basic-setup-script)

## Fix the `download_url`
The release tags have a v prepended to the version.

### DIFF
--- a/source/django/setup.py.erb
+++ b/source/django/setup.py.erb
@@ -3,6 +3,7 @@ from setuptools import setup
 setup(
   name = 'govuk_template_<%= template_abbreviation %>',
   packages = ['govuk_template'],
+  include_package_data=True,
   version = '<%= GovukTemplate::VERSION %>',
   license = 'MIT',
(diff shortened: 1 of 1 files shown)


robyoung:
Ace, thanks. Could you push a rebuilt `0.22.2` python package (ie. [this](https://github.com/alphagov/govuk_template/releases/download/v0.22.2/django_govuk_template-0.22.2.tgz) one) to github releases as well please?

That would mean that we can install straight from the URL.

36degrees:
#320 should take care of that when it's merged.

robyoung:
Ahh, perfect! thanks

PR #323, opened 2017-08-10
Add SRI attributes to templates

h-lame:
For: https://trello.com/c/SVfWxYYs/205-enable-subresource-integrity-sri-on-static-s

We don't actually add the SRI attributes, but instead add hooks to let the template processors themselves use their own api's to add the SRI attributes.  Currently we only expect the rails version of these to work as we our shims for the other processors ignores the integrity options, but they can be filled in once we know how generate integrity values with them.

Resurrects the work from #301 which was removed in #308 when we realised there was a bug in firefox.

### DIFF
--- a/docs/using-with-rails.md
+++ b/docs/using-with-rails.md
@@ -38,3 +38,21 @@ Or to add content to `<head>`, for stylesheets or similar:
 ```
 
 Check out the [full list of blocks](template-blocks.md) you can use to customise the template.
+
+## SRI
+
+`govuk_template` >= 20.0.0 can be used together with `sprockets-rails` >= 3.0.0 in order to make use of the SRI
+
+You can read more about SRI [here](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).
+
+SRI will add an `integrity` attribute on your script tags:
+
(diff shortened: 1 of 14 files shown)


igloosi:
@h-lame how can this be tested?

fofr:
Does this vary much from https://github.com/alphagov/govuk_template/pull/301?

docs/using-with-rails.md
```diff
@@ -38,3 +38,20 @@ Or to add content to `<head>`, for stylesheets or similar:
 ```
 
 Check out the [full list of blocks](template-blocks.md) you can use to customise the template.
+
+## SRI
```
fofr: Do the browser bugs we encountered need mentioning? And the gem you wrote to fix them?

h-lame:
@fofr - the implementation is mildly different (pulling in less of the rails helper methods and inlining some code), but part from that it's the same.

docs/using-with-rails.md
h-lame: Good point, I'll update the readme.

h-lame: done in a rebase.

PR #324, opened 2017-08-29
Making it possible to not include fonts.css

jonheslop:
Some of GOV.UK Pay clients want the ability to have their own branded version of our payments form this means that they don't want NTA either, so rather than do a load of overrides, I'd like to not include the CSS at all (this also reduces HTTP reqs)

I'm haven't written much Ruby nor worked on govuk_template before… but does this look like the right approach about making it optional?

### UPDATE
After speaking with @fofr I cleaned this up and have applied it to all templating languages, how does it look?

### DIFF
--- a/build_tools/compiler/django_processor.rb
+++ b/build_tools/compiler/django_processor.rb
@@ -23,6 +23,7 @@ module Compiler
       footer_top:           block_for(:footer_top),
       homepage_url:         statement_tag_for(:homepage_url, 'https://www.gov.uk'),
       global_header_text:   statement_tag_for(:global_header_text, 'GOV.UK'),
+      fonts_stylesheets:    statement_tag_for(:fonts_stylesheets, '<!--[if IE 8]><link href="{% static \"fonts-ie8.css\" %}" media="all" rel="stylesheet" /><![endif]--><!--[if gte IE 9]><!--><link href="{% static \"fonts.css\" %}" media="all" rel="stylesheet" /><!--<![endif]-->'),
       head:                 block_for(:head),
       header_class:         block_for(:header_class),
       html_lang:            statement_tag_for(:html_lang, 'en'),
--- a/build_tools/compiler/mustache_inheritance_processor.rb
+++ b/build_tools/compiler/mustache_inheritance_processor.rb
@@ -19,6 +19,8 @@ module Compiler
       footer_top:           tag_for(:footerTop),
       homepage_url:         tag_for(:homepageUrl, "https://www.gov.uk/"),
       global_header_text:   tag_for(:globalHeaderText, "GOV.UK"),
+      fonts_stylesheets:    tag_for(:fontsStylesheets, '<!--[if IE 8]><link href="{{{ assetPath }}}stylesheets/fonts-ie8.css?0.22.3" media="all" rel="stylesheet" /><![endif]-->
+          <!--[if gte IE 9]><!--><link href="{{{ assetPath }}}stylesheets/fonts.css?0.22.3" media="all" rel="stylesheet" /><!--<![endif]-->'),
       head:                 tag_for(:head),
       header_class:         tag_for(:headerClass),
       html_lang:            tag_for(:htmlLang, "en"),
--- a/build_tools/packager/mustache_inheritance_packager.rb
+++ b/build_tools/packager/mustache_inheritance_packager.rb
@@ -10,6 +10,26 @@ module Packager
       @base_name = "mustache_inheritance_govuk_template-#{GovukTemplate::VERSION}"
     end
 
+    def build
+      @target_dir = @repo_root.join('pkg', @base_name)
+      @target_dir.rmtree if @target_dir.exist?
+      @target_dir.mkpath
+      Dir.chdir(@target_dir) do |dir|
+        generate_package_json
+        prepare_contents
+        create_tarball
+      end
+    end
+
+    def generate_package_json
+      template_abbreviation = "mustache_inheritance"
+      template_name = "{{ mustache }}"
+      contents = ERB.new(File.read(File.join(@repo_root, "source/package.json.erb"))).result(binding)
+      File.open(File.join(@target_dir, "package.json"), "w") do |f|
+        f.write contents
+      end
+    end
+
     def process_template(file)
       target_dir = @target_dir.join(File.dirname(file))
       target_dir.mkpath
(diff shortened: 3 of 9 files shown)


build_tools/compiler/mustache_inheritance_processor.rb
```diff
@@ -1,6 +1,10 @@
 require 'erb'
 require_relative 'template_processor'
 
+$fontsCSS = '<!--[if IE 8]><link href="{{{ assetPath }}}stylesheets/fonts-ie8.css?0.22.3" media="all" rel="stylesheet" /><![endif]-->
```
jonheslop: This is probably not the right place to store this, anybody have any suggestions?

build_tools/compiler/mustache_inheritance_processor.rb
```diff
@@ -1,6 +1,10 @@
...
+    <!--[if gte IE 9]><!--><link href="{{{ assetPath }}}stylesheets/fonts.css?0.22.3" media="all" rel="stylesheet" /><!--<![endif]-->
+    <!--[if lt IE 9]><script src="{{{ assetPath }}}javascripts/ie.js?0.22.3"></script><![endif]-->'
+
```
jonheslop: This is probably not the right place to store this, anybody have any suggestions?

nickcolley: I don't think that ie.js is related to the fonts.. 🤔

source/views/layouts/govuk_template.html.erb
```diff
@@ -12,9 +12,13 @@
     <!--[if IE 8]><link href="<%= asset_path "govuk-template-ie8.css" %>" media="screen" rel="stylesheet" /><![endif]-->
     <link href="<%= asset_path "govuk-template-print.css" %>" media="print" rel="stylesheet" />
 
+    <% if content_for?(:fonts_css) %>
+      <%= yield(:fonts_css) %>
+    <% else %>
```
nickcolley: I'm not sure this else is needed since we're providing a default?

build_tools/compiler/mustache_inheritance_processor.rb
nickcolley: (Also you could probably get away with having it inline below...)

source/views/layouts/govuk_template.html.erb
jonheslop: Yeah, I tried to use the else before I put that variable in the other file, but this didnt work, whereas that did, should have taken it out, have done now.

build_tools/compiler/mustache_inheritance_processor.rb
jonheslop: Whoops about the IE stuff

jonheslop:
Also I guess, I will have to alter the processors for each flavour and not just the mustache inheritance one… but wanted to check I was doing it right first

build_tools/packager/mustache_inheritance_packager.rb
```diff
@@ -10,6 +10,26 @@ def initialize
       @base_name = "mustache_inheritance_govuk_template-#{GovukTemplate::VERSION}"
     end
 
+    def build
```
nickcolley: What are these methods doing? Why are they needed for this PR?

jonheslop: Oh it looks like someone started work on the mustache inheritance stuff and it never got finished and so was never made into a tarball which means you can include it in your package.json. I just copied that from the mustache one. 

Could do this as a separate PR I guess?

build_tools/compiler/mustache_inheritance_processor.rb
```diff
@@ -19,6 +19,8 @@ def self.tag_for(lowerCamelCaseKey, default_value="")
       footer_top:           tag_for(:footerTop),
       homepage_url:         tag_for(:homepageUrl, "https://www.gov.uk/"),
       global_header_text:   tag_for(:globalHeaderText, "GOV.UK"),
+      fonts_css:            tag_for(:fontsCSS, '<!--[if IE 8]><link href="{{{ assetPath }}}stylesheets/fonts-ie8.css?0.22.3" media="all" rel="stylesheet" /><![endif]-->
```
fofr: Technically this isn't CSS, it's HTML.

source/views/layouts/govuk_template.html.erb
```diff
@@ -12,8 +12,9 @@
     <!--[if IE 8]><link href="<%= asset_path "govuk-template-ie8.css" %>" media="screen" rel="stylesheet" /><![endif]-->
     <link href="<%= asset_path "govuk-template-print.css" %>" media="print" rel="stylesheet" />
 
-    <!--[if IE 8]><link href="<%= asset_path "fonts-ie8.css" %>" media="all" rel="stylesheet" /><![endif]-->
-    <!--[if gte IE 9]><!--><link href="<%= asset_path "fonts.css" %>" media="all" rel="stylesheet" /><!--<![endif]-->
+    <% if content_for?(:fonts_css) %>
```
fofr: As suggested on Slack a while back:

Would it be simpler for this to be `:exclude_new_transport` (or some other name) and use the existing `:head` variable to include any alternative font.

jonheslop: Ah yeah, thanks I knew you'd suggested something but I've been away for two weeks and slack history doesn't go back anywhere near that long 🙄

build_tools/packager/mustache_inheritance_packager.rb
jonheslop: I got this totally backwards (sorry been on holiday for two weeks) it created the tarball but not the uncompressed version (all the other flavours do) which makes it easier to develop

selfthinker:
Could you rewrite your commit messages to meet our [git guidelines](https://github.com/alphagov/styleguides/blob/master/git.md)?

jonheslop:
@selfthinker How does it look now?

build_tools/compiler/django_processor.rb
```diff
@@ -23,6 +23,7 @@ def self.statement_tag_for(key, default_value)
       footer_top:           block_for(:footer_top),
       homepage_url:         statement_tag_for(:homepage_url, 'https://www.gov.uk'),
       global_header_text:   statement_tag_for(:global_header_text, 'GOV.UK'),
+      fonts_stylesheets:    statement_tag_for(:fonts_stylesheets, '<!--[if IE 8]><link href="{% static \"fonts-ie8.css\" %}" media="all" rel="stylesheet" /><![endif]--><!--[if gte IE 9]><!--><link href="{% static \"fonts.css\" %}" media="all" rel="stylesheet" /><!--<![endif]-->'),
```
fofr: There's a lot of repetition of the fonts stylesheet markup. Can you put that into a helper and have all the processors reference the same string.

This will also make bumping the font version easier.

selfthinker:
The commit message looks much better now, thanks. 👍

build_tools/compiler/django_processor.rb
jonheslop: Good point, will do that

timpaul:
Thanks for this work.

PR #326, opened 2017-09-21
Fix #325 Focused, unvisited links are (extremely) low-contrast

### LINKED ISSUE #264: Focused links have low contrast

### LINKED ISSUE #325: Focused, unvisited links are (extremely) low-contrast
An attempt to fix #264 raised this issue of having blue text on green buttons when focused on unvisited links.
screen shot 2017-09-15 at 14 21 40
screen shot 2017-09-15 at 14 21 53

Tested with latest Chrome, Firefox on OSX and Windows and IE9, IE10, IE11 on Windows.

If we do want to [make focused links slightly darker to improve contrast](https://github.com/alphagov/govuk_template/commit/79466a489c38b0b82f4cb95daa2baba41b375a81), I would get rid of `:link` (which selects unvisited links). So the code below
`a:link:focus { color: darken( $link-colour, 2.5%) }`
would become
`a:focus { color: darken( $link-colour, 2.5%) }`
or even simpler, I would move the color attribution inside `a:focus`, which is [a few lines above](https://github.com/alphagov/govuk_template/commit/79466a489c38b0b82f4cb95daa2baba41b375a81#diff-d0d347595f17783ceb1cfde3086eee86R58).

In this way we will have a "less powerful" selector (for improving contrast on links) that won't affect buttons.

If that sounds good, let me know and I'll come up with a PR.

alex-ju:
This commit fixes #325, which is in essence a fix for PR #272 which was an attempt to fix #264.

If that sounds confusing enough, here is the short story:
- Now we have blue text on green buttons when focused on unvisited links.
- This PR is making use of a "less powerful" selector for improving contrast on links while not affecting buttons.

**Buttons**
Before:
screen shot 2017-09-15 at 14 21 53
After:
screen shot 2017-09-21 at 15 57 26

Tested with latest Chrome, Firefox on OSX and Windows and IE9, IE10, IE11 on Windows.

### DIFF
--- a/source/assets/stylesheets/_accessibility.scss
+++ b/source/assets/stylesheets/_accessibility.scss
@@ -58,10 +58,7 @@ a {
 a:focus {
   background-color: $focus-colour;
   outline: 3px solid $focus-colour;
-}
-
-/* Make links slightly darker when focused to improve contrast. */
-a:link:focus {
+  /* Make links slightly darker when focused to improve contrast. */
   color: darken( $link-colour, 2.5%)
 }
 


edwardhorsford:
Is this issue about the blue text on buttons?

I'm not clear if things are fixed from the images - the button 'after' example still looks wrong. It also looks like focused links are now black - was that the intention?

alex-ju:
@edwardhorsford sorry for the misleading screenshots. updated.

edwardhorsford:
Thanks @alex-ju 

Is there a reason focused regular links are now black rather than blue? I don't have a huge opinion on it, but it's a new change (over and above fixing the original bug as I saw it).

alex-ju:
@edwardhorsford you're right, the regular links are not affected by this change; they will only be slightly darker when focused, as before.

nickcolley:
Just ran into this, can we get this in. @edwardhorsford any objections?

edwardhorsford:
@nickcolley I forget what this PR is about. Skimming it, it sounds like the PR was changing things outside of buttons - is that still the case?

36degrees:
I *think* this will mean that visited links, when focussed, will be dark blue rather than purple. How much of an issue is that?

The alternative, I think, would be to make the button more defensive – it would need to set a white text colour against the `:link:focus` and `:visited:focus` pseudo selectors?

alex-ju:
The intention was to make the only [the unvisited links darker](https://github.com/alphagov/govuk_template/pull/272#discussion_r98440877). This PR doesn't fix that, so I'll close for now with the heads up that the problem is still there.

PR #332, opened 2017-11-24
Ensure defaults aren't HTML escaped in Play! processor

### LINKED ISSUE #276: Play template package doesn’t disable HTML escaping for Crown Copyright entity
### Steps to reproduce

Examine the Crown Copyright section of the packaged Play template

### Expected behaviour

The copyright HTML entity is wrapped in a call to `Html()`, to avoid it being HTML-escaped

```
<a href="http://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/copyright-and-re-use/crown-copyright/">@crownCopyrightMessage.getOrElse(Html("&copy; Crown copyright"))</a>
```

### Actual behaviour

The copyright HTML entity is not wrapped in an `Html()` call, resulting in it being HTML-escaped, and thus appearing as `&copy;` in the browser.

```
<a href="http://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/copyright-and-re-use/crown-copyright/">@crownCopyrightMessage.getOrElse("&copy; Crown copyright")</a>
```

escaped

### Workaround

I’m currently working around the bug by duplicating the Crown copyright message content in my main template:

```
@crownCopyrightMessage = {
    &copy; Crown copyright
}
...
@govuk_template(... Some(crownCopyrightMessage))(content)
```

This content isn’t HTML-escaped, so the copyright entity appears correctly when the resulting pages are viewed in a web browser.

TheNickWilson:
Some of the default content contains special characters and markup, which
we do not want to be HTML escaped.

Fixes #276

### DIFF
--- a/build_tools/compiler/play_processor.rb
+++ b/build_tools/compiler/play_processor.rb
@@ -26,11 +26,11 @@ module Compiler
       footer_support_links: '@footerLinks',
       homepage_url: '@homepageUrl.getOrElse("https://www.gov.uk/")',
       global_header_text: '@globalHeaderText.getOrElse("GOV.UK")',
-      cookie_message: '@cookieMessage.getOrElse("<p>GOV.UK uses cookies to make the site simpler. <a href=\"https://www.gov.uk/help/cookies\">Find out more about cookies</a></p>")',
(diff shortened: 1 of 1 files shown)


36degrees:
Thanks for this!

PR #337, opened 2018-01-24
Fix some meta and link tags not being closed

StupidCodeFactory:
**CHANGES:**

While preparing a load test script with JMeter, the XPath parser
would fail because some `link` and `meta` tags weren't closed.


36degrees:
`<link>` and `<meta>` tags are ['void elements'](https://www.w3.org/TR/html5/syntax.html#void-elements) which means they do not need to be closed. XPath only works on 'well formed XML documents', which regular HTML is not (while XHTML is).

Additionally, following the launch of the [GOV.UK Design System](https://design-system.service.gov.uk/), GOV.UK Template will now only get major bug fixes and security patches, so I'm going to close this.

PR #340, opened 2018-03-06
0.24.0 font loading method update and WOFF / WOFF2 versions of fonts

Nooshu:
- Included EOT / WOFF / WOFF2 version of the fonts in a single @font-face rule.
- Removed IE8 version of the fonts from the `<head>`. This now comes from the same rule.
- Added font-display property for browsers that support it.
- Bumped the release to 0.24.0


tijmenb:
I'll try using this branch on GOV.UK so we can observe the impact.

nickcolley:
Hey Matt, this is looking interesting.

Two things:

1. This would have a massive impact to how GOV.UK pages load, can you include more detail on what to expect.
The original approach blocked the page rendering to avoid a flash of text, what compromises does this approach make, what are the pros and cons?
2. There's been some uncertain conversation around licences and if we can have these fonts in a public repo, leading to https://github.com/alphagov/fonts however this has never been consistent due to the EOT files being in the repository, would be good to go for the simplest approach that fits what should be done legally.

PR #343, opened 2018-04-12
Fix proposition menu accessibility

benthorner:
Fix two accessibility issues with proposition links:

   - Clicking on the 'Menu' button while using VoiceOver on a narrow-width device would lock the focus on the 'Menu' button with no way to escape. The button is now always visible to avoid this.
   - The proposition links section was difficult to navigate with VoiceOver due to a lack of aria attributes, the 'Menu' being a link (not a button) and the ordering of the 'Service Name' link.

I've also added some docs to help future manual testing.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -8,6 +8,37 @@ The source files are in the `/source` directory.  The `compile` rake task builds
 
 This resulting app directory is included in the gem and hooked in as a Rails engine
 
+## Getting a manual test rig setup
+
+You will need a 'test website' to manually verify changes. If you're only modifying the existing assets, you can do this with NodeJS.
+
+```
+npm install -g http-server node-sass
+
+# Serve all the assets in the current directory
+http-server -c-1
+
+# Auto-compile all the SASS to CSS
+node-sass --recursive source/assets/stylesheets --output app/assets/stylesheets --include-path ../govuk_frontend_toolkit/stylesheets --watch
+```
+
+Then create a skeleton `test.html` file in the `app` directory (so git ignores it) using the following template.
+
+```
+<html>
+  <head>
+    <link rel="stylesheet" href="assets/css/govuk-template.css"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  </head>
+  <body class="js-enabled">
+    YOUR TEST CODE HERE
+  </body>
+  <script type="text/javascript" src="../source/assets/javascripts/govuk-template.js"></script>
+</html>
+  </body>
+</html>
+```
+
 ## Extra details for the tarball build
 
 The tarball build process takes the compiled template and assets from the `/app` directory, and performs some extra processing:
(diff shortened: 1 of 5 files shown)


docs/development.md
```diff
@@ -8,6 +8,37 @@ The source files are in the `/source` directory.  The `compile` rake task builds
 
 This resulting app directory is included in the gem and hooked in as a Rails engine
 
+## Getting a manual test rig setup
```
nickcolley: I believe there's a rails app to do this instead?

Not sure how http://alphagov.github.io/govuk_template/ is served but I imagine this could be used to test changes.

docs/development.md
```diff
@@ -8,6 +8,37 @@ The source files are in the `/source` directory.  The `compile` rake task builds
...
+
+You will need a 'test website' to manually verify changes. If you're only modifying the existing assets, you can do this with NodeJS.
+
+```
+npm install -g http-server node-sass
+
+# Serve all the assets in the current directory
+http-server -c-1
```
nickcolley: You might want to use 'npx' so that people dont need to install http-server and node-sass globally.

benthorner: Cool - that's a new one for me.

docs/usage.md
```diff
@@ -14,10 +14,10 @@ You can get a propositional title and navigation by setting the content for `hea
 
     <div class="header-proposition">
       <div class="content">
-        <a href="#proposition-links" class="js-header-toggle menu">Menu</a>
+        <a href="/" id="proposition-name">Service Name</a>
+        <button href="#proposition-links" class="js-header-toggle menu" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
```
igloosi: href is not valid `<button>` attribute. we either go back to `<a>` tag or remove that from the button.

source/assets/stylesheets/_header.scss
```diff
@@ -117,13 +117,17 @@
...
       color: $white;
       display: block;
       float: right;
       text-decoration: none;
-      padding-top: 6px;
+      font-weight: bold;
+      padding: 2px 6px 1px;
+      font-size: 13px !important;
+      background-color: $black;
+      border-radius: 3px;
```
igloosi: in relation on the comment above, can we not introduce new visual to the button/a tag?
if we're changing it to a button then let's only add styles that make it look like it did before. 
no border radius please :)

igloosi:
good stuff otherwise @benthorner

docs/usage.md
benthorner: Unfortunately it would make the JS code non-backward compatible to remove it. Instead I've switched back a to a link but changed it's role to button to make it less confusing with VoiceOver.

source/assets/stylesheets/_header.scss
benthorner: So the idea Oliver had was to make the link look like the developer docs, but actually it's a smaller change just to keep the link and change it's role to button for a better VoiceOver.

benthorner: So back to a link and no visual change :-)

docs/usage.md
benthorner: So back to a link.

source/assets/stylesheets/_header.scss
igloosi: thumbs 👍

igloosi:
@benthorner let me know when it's ready to be looked at

PR #347, opened 2018-04-17
Make sure js-hidden class is scoped by js-enabled

benthorner:


source/assets/stylesheets/_accessibility.scss
```diff
@@ -12,9 +12,9 @@
   }
 }
 
-/* Hide for both screenreaders and browsers */
+/* Hide for both screenreaders and browsers when JS enabled */
```
36degrees: This comment only applies to the second selector, is there a better way to comment this?

benthorner: True - we can just lose the JS bit.

PR #349, opened 2018-04-18
Make sure the default opengraph image is the fallback

tijmenb:
Currently when an application sets a custom og:image (which it should), the custom meta tag will be inserted below the default meta tag.

This means that Facebook still sees the big black crown as the primary image and the user has to manually select a new image. Slack doesn't allow you to change the image, so the custom `og:image` is completely ignored.
 
By flipping the meta tag and `:head` inserter, we make sure that the custom image is the first one chosen for the page.

## Effect on Facebook

screen shot 2018-04-18 at 15 27 15

screen shot 2018-04-18 at 15 27 21

## Effect on Slack

screen shot 2018-04-18 at 15 30 18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Make sure that if a custom `og:image` meta tag is set in the `:head` block,
+  this image has priority over the default image.
+
 # 0.23.2
 
 - Fix missing js-hidden class. This class is still used by downstream projects and should be kept for backwards compatibility.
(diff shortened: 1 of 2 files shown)


nickcolley:
Super cool, could you add a comment in the template that briefly explains that it being below the `yield` is important so this doesnt get 'cleaned' up later?

tijmenb:
@nickcolley good point, updated!

nickcolley:
@tijmenb thanks! Once merged you'll need to do a separate pull request to release this: https://github.com/alphagov/govuk_template/pull/348

tijmenb:
@nickcolley thanks! Also updated the changelog.

PR #355, opened 2018-04-27
Add instructions for incorporating font changes

quis:
Don’t feel like this is super-obvious.


Nooshu:
Sounds good Chris, could we quickly change the line:

> Base64 encoded strings to WOFF and EOT files...

To

> Base64 encoded strings to WOFF2, WOFF and EOT files...

Just so were are very explicit around all the filetypes that need to be included.

quis:
@Nooshu see d921746

PR #361, opened 2018-06-22
Update crown copyright link

rubenarakelyan:
This commit updates the crown copyright link which points to a page on The National Archives website, which has moved.

It also updates the gem developer details to make them consistent with our other gems.


nickcolley:
We'll likely want to update https://github.com/alphagov/govuk-frontend/blob/106133d368c65237fe178b3b77a08c355b8809d4/src/components/footer/template.njk#L75 too.

PR #365, opened 2018-07-31
Add deprecation notice and styles for the website

igloosi:
This is to be merged to the `gh-pages` branch
Also my editor is set to remove trailing whitespace....

### DIFF
--- a/assets/stylesheets/govuk-template.css
+++ b/assets/stylesheets/govuk-template.css
@@ -620,3 +620,21 @@ button:focus {
         @media (min-width: 641px) {
           #footer .footer-meta .copyright a {
             text-align: right; } }
+/* deprecation notice */
+.notification-summary {
+  border: 4px solid #005ea5;
+  margin-top: 15px;
+  margin-bottom: 15px;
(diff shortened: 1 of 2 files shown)


joelanman:
Should we say we are no longer supporting other templating languages?

36degrees:
I think these changes will both be overwritten by the [DocsPublisher](https://github.com/alphagov/govuk_template/blob/master/build_tools/publisher/docs_publisher.rb#L35-L48) the next time we do a release.

I believe these changes need to be made to the [`data/index.yml`](https://github.com/alphagov/govuk_template/blob/gh-pages/data/index.yml) file which is used to generate the HTML files.

I think the extra styles should probably be either made inline or added to the style block at the top of that 'data file', as the stylesheet will be regenerated as well.

We'll then need to run the DocsPublisher by itself – annoyingly it looks like that can't be done without creating a release, so we may need to update the Rakefile and .travis.yml to make that possible…

igloosi:
give me strength...

igloosi:
closing, as we'll just go with the notice in the README.md

PR #371, opened 2019-05-01
[WIP][DO NOT MERGE] Don't apply styles to new cookie banner

vanitabarrett:
We'll be testing a new cookie banner. These styles don't play nicely with the new banner, but we can't remove them yet, while the old banner is still being used (as the new banner is still in testing phase). We can instead make them specifically only apply to the new cookie banner so they don't affect the new one while we're testing.

This isn't ideal code, as we're referring to component class selectors outside the component itself. However, this code should be temporary - when the new banner has been made live, we can remove this styling.

## Current ("old") banner
Screen Shot 2019-05-01 at 15 06 44

## New banner (WIP)
Screen Shot 2019-05-01 at 15 07 05

### DIFF
--- a/source/assets/stylesheets/_header.scss
+++ b/source/assets/stylesheets/_header.scss
@@ -240,11 +240,12 @@
   display: none; /* shown with JS, always on for non-JS */
 }
 
-#global-cookie-message {
+.gem-c-cookie-banner#global-cookie-message {
   width: 100%;
   background-color: $light-blue-25;
   padding-top: 10px;
   padding-bottom: 10px;
+
   p {
     @extend %site-width-container;
     @include core-16;


vanitabarrett:
[Build functioning consent cookies banner](https://trello.com/c/5WxS01Xd/120-build-functioning-consent-cookies-banner)

nickcolley:
Seems fine as a temporary thing, but note that this repository is deprecated.

One thing you could consider is the following:

1. Remove the cookie banner from `govuk_template` (given that it doesnt work as intended anymore)
2. Experiment with banner at the GOV.UK Publishing level
3. Propose the banner as a new component to the GOV.UK Design System.

vanitabarrett:
Good point @nickcolley , it might make it easier for us in the long run to move this into the component styles now, so it's easier to work with anyway 👍

PR #373, opened 2019-07-12
Wrap JS code around a document.body loaded function

cintamani:
On IE11 we often fail to run JS code because when this line of code gets executed the `document.body` is not loaded yet. 
This PR then wraps the code around a document ready VanillaJS function.

### DIFF
--- a/source/views/layouts/govuk_template.html.erb
+++ b/source/views/layouts/govuk_template.html.erb
@@ -37,7 +37,11 @@
   </head>
 
   <body<%= content_for?(:body_classes) ? raw(" class=\"#{yield(:body_classes)}\"") : '' %>>
-    <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
+    <script>
+      document.addEventListener("DOMContentLoaded", function() {
+       document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
+      });
+    </script>
 
     <%= yield :body_start %>
 


36degrees:
Hi @cintamani,

Do you have an example of the sorts of errors you're seeing, and/or a URL where you're seeing this behaviour that I can use to try and reproduce this?

This code has been in place for ~6 years, and we are using the same approach in GOV.UK Frontend, but I'm not aware of this issue having come up before.

As I understand it, at the point the `<script>` runs, the `<body>` tag should have been parsed by the browser and be part of the DOM, and so `document.body` should be defined?

This makes me think there might be something more to this – is there any other JavaScript running on your page that is trying to modify the classes on the body tag? Could there be some sort of race condition going on?

cintamani:
Hi @36degrees! So you can reproduce the issue using this live service we went live with recently: https://wasteexemptions.service.gov.uk/ 
Feel free to inspect. It doesn't happen all the times, so you will need to refresh until you see it, while keeping the console open. 
The error is the following: 
image-20190711-095529
From the console is easy to see the line of code causing it, and that's the one I am trying to fix in this PR. 
We use this template through this gem: https://github.com/alphagov/govuk_elements_rails 
Let me know :)

36degrees:
I spent some time looking into this with @cintamani. 

After a bit of digging, we think this might be being caused by their application using [turbolinks](https://github.com/turbolinks/turbolinks) which hijacks the native browser navigation:

> When you follow a link, Turbolinks automatically fetches the page, swaps in its `<body>`, and merges its `<head>`, all without incurring the cost of a full page load.

I think this would be a breaking change, as other services may well have scripts that rely on the js-enabled class existing on the body before the DOMContentLoaded event is fired.

For that reason, I'm going to close this now. If anyone else is seeing this behaviour, and isn't using turbolinks, please open an issue.

